PR #2, opened 2019-02-14
Added packet.MarshalTo() and made packet.Marshal() faster

kixelated:
`packet.MarshalTo()` allows the caller to potentially avoid allocations. The `packet.MarshalSize()` method exists to inform the caller the size of the buffer we're expecting in order to prevent allocations. 

I removed `packet.Header.PayloadOffset` because it can be computed with `packet.Header.MarshalSize()`. I would be fine with `packet.Header.PayloadOffset()` as a function, but I really don't like `Marshal` causing side-effects. It should write the current values, not mutate them.

Along those lines, I also removed `packet.Raw`. You can call `packet.Marshal()` if you want the raw data. There's probably some disorganized code that tried to optimize performance by using this variable but it's dangerous. You could `packet.Marshal()` and then change a parameter but `packet.Raw` wouldn't be updated.

I like this `rtp.Packet` implementation because it only exposes struct variables found in the spec. Micro-performance optimizations need to be done without polluting the API. `payload.Header.PayloadOffset` saves nanoseconds compared to computing the offset, at the cost of a more frail API.

```
BenchmarkMarshalToNew-8   	100000000	        13.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkMarshalNew-8   	30000000	        49.2 ns/op	      32 B/op	       1 allocs/op
BenchmarkMarshalOld-8   	20000000	        89.6 ns/op	      80 B/op	       2 allocs/op
```

As you can see, memory management is HUGE in Go. We see a 2x speed up by avoiding a reallocation, and another 3.5x speed up by avoiding allocating altogether.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ Check out the **[contributing wiki](https://github.com/pions/webrtc/wiki/Contrib
 * [Sean DuBois](https://github.com/Sean-Der) - *Original Author*
 * [Woodrow Douglass](https://github.com/wdouglass) *RTCP, RTP improvements, G.722 support, Bugfixes*
 * [Michael MacDonald](https://github.com/mjmac)
+* [Luke Curley](https://github.com/kixelated) *Performance*
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/packet.go
+++ b/packet.go
@@ -7,16 +7,12 @@ import (
 	"github.com/pkg/errors"
 )
 
-// TODO(@kixelated) Remove Header.PayloadOffset and Packet.Raw
-
 // Header represents an RTP packet header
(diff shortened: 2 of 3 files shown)


kixelated:
@Sean-Der I decided that I'd like to remove `Packet.Raw` and `Header.PayloadOffset`. I like that it will cause a compilation error if anybody actually uses the fields, and that there are drop-in replacements.

I'm not 100% sure if `MarshalSize()` is the best name and if it should be exported. But it's consistent, makes sense, and it's needed to replace `Header.PayloadOffset`. Let me know if you have any suggestions!

kixelated:
replaced by #3 in order to switch repos

PR #3, opened 2019-02-15
Added packet.MarshalTo() and made packet.Marshal() faster

kixelated:
`packet.MarshalTo()` allows the caller to potentially avoid allocations. The `packet.MarshalSize()` method exists to inform the caller the size of the buffer we're expecting in order to prevent allocations. 

I removed `packet.Header.PayloadOffset` because it can be computed with `packet.Header.MarshalSize()`. I would be fine with `packet.Header.PayloadOffset()` as a function, but I really don't like `Marshal` causing side-effects. It should write the current values, not mutate them.

Along those lines, I also removed `packet.Raw`. You can call `packet.Marshal()` if you want the raw data. There's probably some disorganized code that tried to optimize performance by using this variable but it's dangerous. You could `packet.Marshal()` and then change a parameter but `packet.Raw` wouldn't be updated.

I like this `rtp.Packet` implementation because it only exposes struct variables found in the spec. Micro-performance optimizations need to be done without polluting the API. `payload.Header.PayloadOffset` saves nanoseconds compared to computing the offset, at the cost of a more frail API.

```
BenchmarkMarshalToNew-8   	100000000	        13.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkMarshalNew-8   	30000000	        49.2 ns/op	      32 B/op	       1 allocs/op
BenchmarkMarshalOld-8   	20000000	        89.6 ns/op	      80 B/op	       2 allocs/op
```

As you can see, memory management is HUGE in Go. We see a 2x speed up by avoiding a reallocation, and another 3.5x speed up by avoiding allocating altogether.


kixelated:
Okay, while digging into why this change causes `pions/webrtc` unit tests to hang, I found a bug. `Header.PayloadOffset` has a different value depending on if `Marshal` or `Unmarshal` was used.

For `Unmarshal` it will be set to the offset after the optional extension header/payload.
For `Marshal` it will be set to the offset before the optional extension header/payload.

I'm not sure which is right. There's also a 3rd option where the offset should be between the extension header and payload, if you're actually supposed to encrypt the extension payload. It's late so I'm not going to bother reading the spec juuuust yet.


At least I'm feeling pretty vindicated trying to remove the field...

kixelated:
Ugh, that didn't fix the webrtc tests from hanging. I'm thinking maybe the old code

kixelated:
Fixed.

packet.go
```diff
@@ -157,61 +168,99 @@ func (h *Header) Marshal() ([]byte, error) {
...
 	}
-	rawHeader[1] |= h.PayloadType
 
-	binary.BigEndian.PutUint16(rawHeader[seqNumOffset:], h.SequenceNumber)
-	binary.BigEndian.PutUint32(rawHeader[timestampOffset:], h.Timestamp)
-	binary.BigEndian.PutUint32(rawHeader[ssrcOffset:], h.SSRC)
-
-	for i, csrc := range h.CSRC {
-		binary.BigEndian.PutUint32(rawHeader[csrcOffset+(i*csrcLength):], csrc)
+	buf = append(buf,
```
Sean-Der: If there was a bug in here (not saying there is) I would have a really hard time fixing/maintaining this.

Is it not possible to keep the explicit old style of doing things?

kixelated: The old style requires calling `MarshalSize()` again and expanding the buffer if needed, which I guess doesn't actually take much time. It also means keeping track of the offset in a bunch of places.

I actually think this code is way more readable and verifiable. It removes the slice offsetting and just deals with a byte at a time. The encoding might seem scary, but it's how [binary.BigEndian](https://golang.org/src/encoding/binary/binary.go#L115) works. I think `rawHeader[csrcOffset+(i*csrcLength):]` is waaaay scarier.

I can change it but I'd like you to reconsider. :D

Sean-Der: The nice thing about using binary.BigEndian is I don't have to worry about a typo, when reviewing this code I have to think about it a lot more. What you are adding now works, but in the future someone might come and make a mistake and it will be harder to catch.

I agree that working with offsets is frustrating also, it would be nice to find the best mix of both worlds.

I would add @backkem @wdouglass @maxhawkins or @kc5nra they all are good bit twiddlers and happy to be overridden if majority disagrees!

maxhawkins: Doesn't this code fail on a big-endian machine?

maxhawkins: I'm agnostic here. If we have good enough tests either should be OK.

The append way means we have to keep track of the offset in fewer places which is a plus. The `binary` package way makes it harder to mix up little endian and big endian.

wdouglass: This looks ok to me, the shifts should make it endian agnostic. I agree that the big append is less scary then the offset additions.

If there's a real performance benefit then that's enough for me, I vote merge it.

wdouglass: P.s. I agree with @maxhawkins... Make sure this is covered in good tests!

kixelated: The performance difference between using `append` and `binary.BigEndian` is minimal to non-existent. Really I just wanted to avoid manually keeping track of offsets, instead relying on `append` to keep track of the offset aka the slice `len`.

If you guys think the shifts are scarier than the old code with `offset += <math>` all over the place, then yeah I can do it the other way.

wdouglass: Maybe we can avoid the shifts *and* the offset math by using bytes.Buffer with binary.BigEndian?

kixelated: It's going to be a lot slower, especially because `binary.BigEndian.Write` uses reflection.

I pushed a commit that uses `binary.BigEndian.PutX`. It's 12% slower than the append version and I think the code is more involved. Using the `offset` const values instead of `offset += ` has  basically zero performance difference.

wdouglass: Honestly, I think the append implementation is worth the 12% speedup in return for a couple of ugly shifts. If this were C i'd hide the shifts behind a macro, but this isn't C.

Give me a few minutes to review the rest of the PR; i'll let you choose whether to merge the append version or the new PutX version.

kixelated:
Pushed a compromise for you guys.

PR #4, opened 2019-02-15
Initialized as a module

kixelated:
Required to use go.mod's `replace` directive for local development.


Sean-Der:
Nice, would you mind dropping the `errors` package also in this commit?

Just replace all the calls to `errors.Errorf` -> `fmt.Errorf` just nice to have 0 dependencies to worry about

PR #9, opened 2019-02-20
Add fuzzer

maxhawkins:
This will help us find crashing bugs in the Marhsal / Unmarhsal functions.


backkem:
Maybe we should come up with some conventions around fuzzing. It seems even this repo may end up having multiple fuzzers. I'd suggest something similar to the `go test` conventions:
- Name the file `target_fuzz.go`, e.g.: `packet_fuzz.go` in this case.
- Name the working directory `fuzzer_target`, e.g.: `fuzzer_packet.go` in this case.
Thoughts @maxhawkins?

maxhawkins:
Good point. I think I need to experiment with the `go-fuzz-build -f` flag.

Another question is where to put the corpus files. For SDP I have a 32m corpus inherited from Chrome's. Is git-lfs enabled on this repo?

backkem:
Keeping these libraries easily go-gettable seems important. Maybe we can make a separate repo (E.g.: pions/fuzz) to hold the corpus(es) to avoid bloating library downloads? The fuzzer could be instrumented to check this repo out.

PR #10, opened 2019-02-21
Optimize h.264 NAL splitting

kixelated:
It's not a bottleneck, I just thought there was a bug and then got
carried away. It's probably faster on larger NALUs.

As compared to `emitNalu`:
```
name        old time/op  new time/op  delta
nextNALU-8  40.1ns ± 0%  22.6ns ± 1%  -43.48%
```

### DIFF
--- a/codecs/h264_packet_test.go
+++ b/codecs/h264_packet_test.go
@@ -2,48 +2,79 @@ package codecs
 
 import (
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
 
 func TestH264Payloader_Payload(t *testing.T) {
+	assert := assert.New(t)
+
 	pck := H264Payloader{}
 	payload := []byte{0x90, 0x90, 0x90}
 
 	// Positive MTU, nil payload
 	res := pck.Payload(1, nil)
-	if len(res) != 0 {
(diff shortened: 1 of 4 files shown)


codecs/h264_packet_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	input := []byte{
+		0x00, 0x00, 0x00, 0x01, 0x67, 0x42, 0x00, 0x1f, 0x01,
+		0x00, 0x00, 0x01, 0x80, 0x00, 0x00, 0x03, 0x00, 0x7a,
+		0x00, 0x00, 0x00, 0x01, 0xab, 0xbc, 0xef,
+	}
+
+	// The first NALU is empty to trim off the start code.
+	nalu, input := nextNALU(input)
+	assert.Len(nalu, 0)
```
maxhawkins: Personal opinion: I don't think test helpers are that helpful.

```go
if len(nalu) != 0 {
  t.Fatalf("nextNalu() returned empty result")
}
```

Is more readable and doesn't need an extra dependency.

Sean-Der:
Mass closing things with over 1 year of inactivity. If anyone is interested this looks like a really fantastic change, probably just need to rebase and should be fine :)

Issue #1, opened 2018-12-26
crasher in codecs.VP8Packet


maxhawkins:
I encountered this out of bounds panic while parsing a malformed VP8 packet.

```
panic: runtime error: slice bounds out of range

goroutine 35 [running]:
github.com/pions/webrtc/pkg/rtp/codecs.(*VP8Packet).Unmarshal(0xc0001c0060, 0xc0015017a0, 0x567, 0xd80, 0x1035, 0xc001ae6c00, 0xace)
        /Users/max/go/pkg/mod/github.com/pions/webrtc@v1.2.0/pkg/rtp/codecs/vp8_packet.go:115 +0x18e
github.com/pions/webrtc/pkg/media/samplebuilder.(*SampleBuilder).buildSample(0xc0012da000, 0xc0001f10e3, 0xc0000a8390)
        /Users/max/go/pkg/mod/github.com/pions/webrtc@v1.2.0/pkg/media/samplebuilder/samplebuilder.go:63 +0x103
github.com/pions/webrtc/pkg/media/samplebuilder.(*SampleBuilder).Pop(0xc0012da000, 0xc000229e58)
        /Users/max/go/pkg/mod/github.com/pions/webrtc@v1.2.0/pkg/media/samplebuilder/samplebuilder.go:101 +0x87
main.(*Demo).handleTrack(0xc000136000, 0x4465980, 0xc0000a8380, 0xc0001e0420)
        /Users/max/projects/asciirtc/main.go:171 +0x5d2
main.(*Demo).Match.func2.1()
        /Users/max/projects/asciirtc/main.go:93 +0x45
sync.(*Once).Do(0xc0001de030, 0xc0001c7f78)
        /usr/local/Cellar/go/1.11.2/libexec/src/sync/once.go:44 +0xb3
main.(*Demo).Match.func2(0xc0001e0420)
        /Users/max/projects/asciirtc/main.go:92 +0x7b
github.com/pions/webrtc.(*RTCPeerConnection).onTrack.func1(0xc0001e8030, 0xc0001e0420, 0xc000144008)
        /Users/max/go/pkg/mod/github.com/pions/webrtc@v1.2.0/rtcpeerconnection.go:300 +0x30
created by github.com/pions/webrtc.(*RTCPeerConnection).onTrack
        /Users/max/go/pkg/mod/github.com/pions/webrtc@v1.2.0/rtcpeerconnection.go:299 +0xef
```

PR #11, opened 2019-02-26
Checks on VP8 codecs Unmarshal

Antonito:
This is still a work in progress

We should implement basic checks on each Unmarshal function such has:
- the payload size is at least equal to the header's size
- check the PayloadType of the RTP packet
- additional checks to prevent any BoF
What do you think ? :)

Should fix https://github.com/pions/rtp/issues/1 and IVF unit tests on https://github.com/pions/webrtc/pull/445

### DIFF
--- a/codecs/vp8_packet.go
+++ b/codecs/vp8_packet.go
@@ -1,6 +1,10 @@
 package codecs
 
-import "github.com/pions/rtp"
+import (
+	"fmt"
+
+	"github.com/pions/rtp"
+)
 
 // VP8Payloader payloads VP8 packets
 type VP8Payloader struct{}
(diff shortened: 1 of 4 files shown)


Sean-Der:
This is a 👍 from me!

Mind adding a simple test, and yourself to the `README.md` ? Then CI should go green
I would also add `Relates to #1` in your commit message, then it will auto link to the issue.

codecs/vp8_packet.go
```diff
@@ -79,6 +83,11 @@ type VP8Packet struct {
 func (p *VP8Packet) Unmarshal(packet *rtp.Packet) ([]byte, error) {
 	payload := packet.Payload
 
+	// If lower than the required header size, malformed header
+	if len(payload) < 4 {
+		return nil, fmt.Errorf("empty payload")
```
Sean-Der: I would delete the comment and put it in the errorf `Payload is not large enough to container header`

codecs/vp8_packet_test.go
```diff
@@ -0,0 +1,94 @@
+package codecs
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/pions/rtp"
+	"github.com/stretchr/testify/assert"
```
Sean-Der: Sorry mind dropping the dependency :( when users import it will get added to their modules (so trying to avoid adding things)

PR #14, opened 2019-03-08
Switch MarshalTo from append to write

kixelated:
This is a lot more consistent with the stdlib when dealing with lower
level byte slices. The idea is that the code will error if the buffer is
too small, rather than silently append. The `MarshalSize` method allows
you to know the exact size required so there's no reason why you can't
allocate the right amount of memory.


kixelated:
And yeah, this is me back-tracking from the original code review where I wanted to append while you guys wanted to write. The big difference is that we check the bounds and never append.

PR #16, opened 2019-03-26
add optional abs-send-time support to Packetizer

wdouglass:
abs-send-time adds a wall-clock timestamp to the end of each video frame (when it's enabled)

It seems like libwebrtc likes these extra timestamps, lets enable it.

### DIFF
--- a/packetizer.go
+++ b/packetizer.go
@@ -17,14 +17,17 @@ type Packetizer interface {
 }
 
 type packetizer struct {
-	MTU         int
-	PayloadType uint8
-	SSRC        uint32
-	Payloader   Payloader
-	Sequencer   Sequencer
-	Timestamp   uint32
-	ClockRate   uint32
-	AbsSendTime int //if this is nonzero, then it's the extension number that's applied to http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
+	MTU              int
+	PayloadType      uint8
+	SSRC             uint32
+	Payloader        Payloader
+	Sequencer        Sequencer
+	Timestamp        uint32
+	ClockRate        uint32
+	extensionNumbers struct { //put extension numbers in here. If they're 0, the extension is disabled (0 is not a legal extension number)
+		AbsSendTime int //http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
+	}
+	timegen func() time.Time
 }
 
 // NewPacketizer returns a new instance of a Packetizer for a specific payloader
@@ -40,11 +43,12 @@ func NewPacketizer(mtu int, pt uint8, ssrc uint32, payloader Payloader, sequence
 		Sequencer:   sequencer,
 		Timestamp:   r.Uint32(),
 		ClockRate:   clockRate,
+		timegen:     time.Now,
 	}
 }
 
 func (p *packetizer) EnableAbsSendTime(value int) {
-	p.AbsSendTime = value
+	p.extensionNumbers.AbsSendTime = value
 }
 
 func toNtpTime(t time.Time) uint64 {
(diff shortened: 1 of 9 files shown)


packetizer.go
```diff
@@ -68,5 +88,27 @@ func (p *packetizer) Packetize(payload []byte, samples uint32) []*Packet {
 	}
 	p.Timestamp += samples
 
+	if p.AbsSendTime != 0 {
+		t := toNtpTime(time.Now()) >> 14
```
maxhawkins: Go doesn't let you stub out time.Now() for tests. This code would be more testable if there was a way to override it here.

packetizer.go
```diff
@@ -68,5 +88,27 @@ func (p *packetizer) Packetize(payload []byte, samples uint32) []*Packet {
...
+		//apply http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
+		packets[len(payloads)-1].Header.Extension = true
```
maxhawkins: What if `len(payloads) == 0`?

packetizer.go
```diff
@@ -68,5 +88,27 @@ func (p *packetizer) Packetize(payload []byte, samples uint32) []*Packet {
...
+		packets[len(payloads)-1].ExtensionProfile = 0xBEDE
+		packets[len(payloads)-1].ExtensionPayload = []byte{
+			//the first byte is
+			// 0 1 2 3 4 5 6 7
+			//+-+-+-+-+-+-+-+-+
+			//|  ID   |  len  |
+			//+-+-+-+-+-+-+-+-+
+			//per RFC65285
```
maxhawkins: Do you have a link for the RFC? I couldnt find it

packetizer.go
```diff
@@ -41,6 +43,24 @@ func NewPacketizer(mtu int, pt uint8, ssrc uint32, payloader Payloader, sequence
...
+func (p *packetizer) EnableAbsSendTime(value int) {
+	p.AbsSendTime = value
+}
+
+func toNtpTime(t time.Time) uint64 {
+	var s uint64
+	var f uint64
+	u := uint64(t.UnixNano())
+	s = u / 1e9
+	s += 0x83AA7E80 //offset in seconds between unix epoch and ntp epoch
```
maxhawkins: does this deal with the 2038 problem?

packetizer.go
```diff
@@ -68,5 +88,27 @@ func (p *packetizer) Packetize(payload []byte, samples uint32) []*Packet {
...
+		packets[len(payloads)-1].ExtensionPayload = []byte{
+			//the first byte is
+			// 0 1 2 3 4 5 6 7
+			//+-+-+-+-+-+-+-+-+
+			//|  ID   |  len  |
+			//+-+-+-+-+-+-+-+-+
+			//per RFC65285
+			//Len is the number of bytes in the extension - 1
+
+			byte((p.AbsSendTime << 4) | 2),
```
maxhawkins: I'm really confused about what `p.AbsSendTime` represents. Can you choose a different name that is more descriptive?

packetizer.go
```diff
@@ -41,6 +43,24 @@ func NewPacketizer(mtu int, pt uint8, ssrc uint32, payloader Payloader, sequence
 	}
 }
 
+func (p *packetizer) EnableAbsSendTime(value int) {
+	p.AbsSendTime = value
```
maxhawkins: Is this necessary if you can just set it directly on the struct?

packetizer.go
wdouglass: How about a function object in the Packetizer structure? normally it'll call time.Now, but inside the package tests can override it?

packetizer.go
wdouglass: the RFC number is a typo, it's 5285. sorry about that.

https://tools.ietf.org/html/rfc5285

packetizer.go
wdouglass: good catch, fixing.

packetizer.go
wdouglass: t.UnixNano will roll over in 2554. the NTP format will roll over sooner (it rolls over every 136 years). 
The timestamp we're deriving from it (abs-send-time) rolls over every 64 seconds -- i don't think it's a problem.

packetizer.go
wdouglass: it's the rtp extension number. I wonder if it should be a sub-structure -- `ExtNumbers.AbsSendTime`?

packetizer.go
wdouglass: Not really, you're right.

wdouglass: wait-- I added it to the interface. `packetizer` isn't exposed.

wdouglass:
> I'm still confused about what this value does. Is it a timestamp? An identifier?

I assume you mean `packetizer.AbsSendTime` (which has been changed to `packetizer.extensionNumbers.AbsSendTime`)

That's the extension number that is declared by the `a=extmap:` line in the sdp; much like `rtpmap`, `extmap` maps a set of identifiers to RTP header extensions.

packetizer.go
maxhawkins: Hmm wonder why Packetizer is an interface in the first place...

Question for a future pull request.

maxhawkins:
Oh! I understand now. It's the mapping from the SDP to the header extension. Thanks for bearing with me I hadn't read RFC 5285.

The spec says "The 8-bit ID is the local identifier of this element", so maybe calling it an ID would make sense? Something like `AbsSendTimeExtID` maybe. I was really confused because I thought it was the actual timestamp not an identifier.

maxhawkins:
As an aside:

Maybe `rtp.Packet` should directly support RFC 5285? Seems you should be able to add an abs-sendtime to a packet without needing to go through the packetizer. We're also going to need to read the abs-sendtime packets at some point. It would be convenient if Pion parsed them too.

packetizer.go
maxhawkins: I liked the flat version better, personally. Sub-structure might make sense if we supported more extensions, but at that point an array makes more sense.

wdouglass:
the reason I added the abs-send-time support to the Packetizer instead of to `rtp.Packet` is because the standard (or what google considers a standard...) dictates that the abs-send-time extension should only be present on the last packet of a fragmented frame. The Packetizer was the easiest place to implement that; `rtp.Packet` already supports RFC5285 as well as it can, but the specific extensions will need to be hooked in at various places

The only way I could see to move more code into the Packet is to make a `Packet.AddAbsSendTimeExtension` call and move the construction of the header there...

maxhawkins:
Just thinking aloud on RFC 5285. Moved my thoughts here: https://github.com/pions/rtp/issues/17

PR looks good otherwise. I am still confused by the name `packetizer.AbsSendTime` but I'm not going to block you if you want to merge it.

Thanks for implementing this!

PR #20, opened 2019-05-19
Set OpusPacket Payload on Unmarshall

raphaelpereira:
Opus packet umarshal was ignoring payload


Sean-Der:
Thank you so much @raphaelpereira 

Would you mind
* Adding yourself to the `README.md`
* Update your commit message so it has a Subject+Body (with an empty line between them)

After that I can merge and will tag a new version and bump `pion/webrtc`. Thanks for using Pion :)

PR #21, opened 2019-05-19
Set OpusPacket Payload on Unmarshall

raphaelpereira:
OpusPacket Unmarshall function was not setting payload


Sean-Der:
Perfect! Thank you so much, when I am at my desk will tag as a new version :)

Thanks for using Pion! If you have anything that can make the project better I would love to hear. I am always available on Slack:)

Issue #17, opened 2019-03-30
Parse RFC 5285 extensions


maxhawkins:
## Summary

Implement [RFC 5285](https://tools.ietf.org/html/rfc5285) RTP header extensions directly in rtp.Packet.

## Motivation

This will allow us to read and write RFC 5285 formatted extensions like abs-sendtime, rtp-audio-level, and video-orientation.

maxhawkins:
It might make sense to get rid of the `Header bool` flag at the same time. If the Header struct contains an extension then Marshal should set it automatically.

maxhawkins:
Sketching an API here:

```go
p := Packet {
	Header: Header{
		Extensions: []Extension {
			{
				ID: 1,
				Payload: payload1,
			},
			{
				ID: 2,
				Payload: payload2,
			},
		},
	},
}
```

Something like this would work for RFC 5285 packets. Not yet sure what to do for packets with a different ExtensionProfile...

maxhawkins:
Maybe useful to add a helper function like `GetExtension(uint8)` also

```go
var a AbsSendTime
payload := pkt.GetExtension(1)
a.Unmarshal(payload)
```

maxhawkins:
Because the identifier 0 is reserved in RFC 5285 maybe we can use it for packets with a different extension profile.

```go
p := Packet {
	Header: Header{
		ExtensionProfile: 0x1234, // not 0xBEDE
		Extensions: []Extension {
			{
				ID: 0,
				Payload: payload1,
			},
		},
	},
}
```

If you want to write a packet with the ExtensionProfile 0x1234 just put the payload in the Extensions array with ID == 0.

maxhawkins:
On second thought, maybe better to use a map:

```go
type Extensions map[int][]byte

type Header struct {
	...
	Extension Extensions
}

p := Packet {
	Header: Header{
		Extensions: Extensions {
			1: payload1,
			2: payload2,
		},
	},
}
```

One caveat: I'm not sure if it's defined in the spec what happens if two extension payloads are returned with the same ID. Need to look into that before going this direction.

maxhawkins:
[Section 4.1](https://tools.ietf.org/html/rfc5285#section-4.1)
```
Each distinct extension MUST have a unique ID.
```

I think that means we can go ahead with the map-based API. Each extension id can only be used once.

maxhawkins:
A `SetExtension()` counterpart to `GetExtension()` would be nice so you don't have to reach into the Header.

```go
[...]

PR #30, opened 2020-01-05
Add transport-cc extension

adwpc:
Add transport-cc extension

#### Reference issue
https://tools.ietf.org/html/draft-holmer-rmcat-transport-wide-cc-extensions-01#page-3

### DIFF
--- a/transportccextension.go
+++ b/transportccextension.go
@@ -1,0 +1,46 @@
+package rtp
+
+import (
+	"encoding/binary"
+	"errors"
+)
+
+const (
+	transportCCExtensionSize = 4
+)
+
+var (
+	errTooSmall = errors.New("buffer too small")
+)
+
+// TransportCCExtension is a extension payload format in
+// https://tools.ietf.org/html/draft-holmer-rmcat-transport-wide-cc-extensions-01
+// 0                   1                   2                   3
+// 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |       0xBE    |    0xDE       |           length=1            |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |  ID   | L=1   |transport-wide sequence number | zero padding  |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+type TransportCCExtension struct {
+	ID                uint8
+	TransportSequence uint16
+}
+
+// Marshal serializes the members to buffer
+func (t *TransportCCExtension) Marshal() ([]byte, error) {
+	buf := make([]byte, transportCCExtensionSize)
+	buf[0] = t.ID<<4&0xf0 + 1
+	binary.BigEndian.PutUint16(buf[1:3], t.TransportSequence)
+	return buf, nil
+}
+
+// Unmarshal parses the passed byte slice and stores the result in the members
+func (t *TransportCCExtension) Unmarshal(rawData []byte) error {
+	if len(rawData) < transportCCExtensionSize {
+		return errTooSmall
+	}
+	t.ID = rawData[0] >> 4
+	t.TransportSequence = binary.BigEndian.Uint16(rawData[1:3])
+	return nil
+}
--- a/transportccextension_test.go
+++ b/transportccextension_test.go
@@ -1,0 +1,63 @@
+package rtp
+
+import (
+	"bytes"
+	"testing"
+)
+
+func TestTransportCCExtensionTooSmall(t *testing.T) {
+	t1 := TransportCCExtension{}
+
+	rawData := []byte{}
+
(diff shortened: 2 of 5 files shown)


transportccextension.go
```diff
@@ -0,0 +1,49 @@
...
+// 0                   1                   2                   3
+// 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |       0xBE    |    0xDE       |           length=1            |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |  ID   | L=1   |transport-wide sequence number | zero padding  |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+type TransportCCExtension struct {
+	ID  uint8
+	L   uint8
```
maxhawkins: Since L will always equal 1 do you need to set it here?

You could just assert that it's 1 on Unmarshal instead. That will make constructing a TransportCCExtenstion struct easier since you won't need to remember to set L: 1

transportccextension.go
```diff
@@ -0,0 +1,49 @@
...
+	TSN uint16
```
maxhawkins: In general ive tried to avoid using acronyms in structs that require a reader to dig back into an RFC to understand what a packet does.

Could this be TransportSequence or something?

transportccextension_test.go
```diff
@@ -0,0 +1,33 @@
...
+		TSN: 2,
+	}
+
+	if t1 != t2 {
+		t.Error("Unmarshal failed")
+	}
+
+	dstData, _ := t2.Marshal()
+	if !bytes.Equal(dstData, rawData) {
+		t.Error("Marshal failed")
```
maxhawkins: Could you add a case testing an empty packet returns errTooSmall?

Maybe also worth testing a packet that has extra bytes at the end parses correctly.

transportccextension.go
```diff
@@ -0,0 +1,49 @@
...
+	// TransportCCExtensionSize 4 byte
+	TransportCCExtensionSize = 4
+)
+
+var (
+	errTooSmall = errors.New("buffer too small")
+)
+
+// TransportCCExtension is a extension payload format in
+// https://tools.ietf.org/html/draft-holmer-rmcat-transport-wide-cc-extensions-01
```
maxhawkins: Will this format weird in the godoc?

The byte layout is important to have but I'm not sure it belongs in tbe docstring

maxhawkins:
Left you some feedback!

I haven't contributed to the project in a while so I don't want to approve until Sean or someone takes a look, but overall it looks good.

transportccextension.go
at-wat: `L` is a `(length-1)` of the element, in this case two bytes.
`L=3` version of Transport CC extension with extra fields looks been proposed.
https://webrtc.org/experiments/rtp-hdrext/transport-wide-cc-02/
```
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |  ID   | L=3   |transport-wide sequence number |T|  seq count  |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |seq count cont.|
 +-+-+-+-+-+-+-+-+
```

In my understanding, rfc5285 defines just a way to transmit some pairs of ID and value which are "defined by profile".
So, I'd like to suggest to make it something general like `HeaderExtension` and marshal/unmarshal between binary and `map[ID][]byte`.
How do you think?

maxhawkins: I've done some thinking along those lines here: https://github.com/pion/rtp/issues/17

Sadly I don't have time to implement it.

I figured we could add that later in a backwards compatible way but do this for now since adam needs this extension immediately.

adwpc: > Looks nice to have header extensions parser!
> I roughly read rfc5285 and added a comment [#30 (comment)](https://github.com/pion/rtp/pull/30#discussion_r363129755).
> Please let me know if my understanding is wrong :)

New RFC:
https://tools.ietf.org/html/rfc8285

adwpc:
> What's this for? LGTM otherwise

packetizer_test.go import

transportccextension.go
```diff
@@ -0,0 +1,47 @@
+package rtp
+
+import (
+	"encoding/binary"
+	"errors"
+)
+
+const (
+	// TransportCCExtensionSize 4 byte
+	TransportCCExtensionSize = 4
```
Sean-Der: Does this need to be exported?

Sean-Der:
@adwpc LGTM! Thanks for grabbing, would you mind adding the tests max mentioned and squashing into one commit. Then happy to approve :) sorry I am slow on this.

transportccextension.go
adwpc: Fixed

adwpc:
> @adwpc LGTM! Thanks for grabbing, would you mind adding the tests max mentioned and squashing into one commit. Then happy to approve :) sorry I am slow on this. I had to read the RFC myself, never implemented any of the congestion-control algos myself.

Added test and one commit

Issue #54, opened 2020-05-02
audio level support


adwpc:
https://tools.ietf.org/html/rfc6464

PR #55, opened 2020-05-02
Support Audio Level Indication

tarrencev:
Adds support for parsing the audio level extension header as described in https://tools.ietf.org/html/rfc6464.

Supports decoding both one and two byte formats but only encodes one byte. Let me know if we would like to support marshaling into two byte as well. In that case let me know what interface would make sense.

#54


tarrencev:
Updated PR to follow the chromium implementation. A little simpler :)

PR #58, opened 2020-05-14
Fix rtp header extensions marshaling/unmarshaling

### LINKED ISSUE #57: Wrong rtp header extension unmarshaling/marshaling
### Your environment.
  - Version: master

Unfortunately PR #56 doesn't correctly Unmarshal and Marshal header extensions since the length value calculation is wrong. 

The code considers it the number of extensions, instead it should be the number of 32-bit words of the total extensions size.

Unfortunately the tests work because they always have three extensions with length=3 or the raw packet is wrong.

```
    0                   1                   2                   3
    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |      defined by profile       |           length              |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |                        header extension                       |
   |                             ....                              |

   If the X bit in the RTP header is one, a variable-length header
   extension MUST be appended to the RTP header, following the CSRC list
   if present.  The header extension contains a 16-bit length field that
   counts the number of 32-bit words in the extension
```

I'm opening a PR to fix this and add more tests.

sgotti:
#### Description

* The length in the RTP Header extension is the number of 32 bit words in the
  extensions. I.E. with rfc 8285 multiple extensions is the size of the all the
  extensions divided by 4.
* The extensions must be be padded to 32-bit boundaries.

* Fix the tests to have a correct raw packet.
* Add some more tests.

#### Reference issue
Fixes #57

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [adwpc](https://github.com/adwpc) *add transport-cc extension*
 * [Bao Nguyen](https://github.com/sysbot) *add VP9 noop, bug fixes.
 * [Tarrence van As](https://github.com/tarrencev) *add audio level extension*
+* [Simone Gotti](https://github.com/sgotti)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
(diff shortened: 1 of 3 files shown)


sgotti:
@tarrencev Thanks! Fixing the failing tests (pushed an old version).

PR #60, opened 2020-05-15
RTP Packet del helper + fix

tarrencev:
Some minor fixes/updates:
* DelExtension helper function
* Return nil in GetExtension if h.Extension is false
* Fix extension size check in SetExtension

### DIFF
--- a/packet_test.go
+++ b/packet_test.go
@@ -447,6 +447,102 @@ func TestRFC8285TwoByteMultipleExtensionsWithLargeExtension(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


packet_test.go
```diff
@@ -447,6 +447,102 @@ func TestRFC8285TwoByteMultipleExtensionsWithLargeExtension(t *testing.T) {
 	}
 }
 
+func TestRFC8285GetExtensionReturnsErrorWhenExtensionsDisabled(t *testing.T) {
```
sgotti, line 450: Shouldn't it be `ReturnsNil`?

PR #56, opened 2020-05-11
RFC 5285 header extension support

tarrencev:
An initial implementation of RFC 5285 header extensions based on previous work in #17. Please provide feedback :)

This is a breaking change unfortunately. We could probably avoid that with some added complexity if we'd like to.

PR #61, opened 2020-05-18
Fix extension padding on MarshalTo to dirty buffer

at-wat:
MarshalTo to non-zero buffer didn't clear padding bytes.
(regression from 4b1b5825)

### Reference issue
Ref #56 and #58


sgotti:
@at-wat LGTM! I totally forgot that the buf could be reused. Thanks!

PR #62, opened 2020-05-18
Fix AbsSendTime marshal/unmarshal in Packetizer

at-wat:
ID/len byte of AbsSendTime was duplicated
and ID of AbsSendTime extension was always unmarshalled as 0.
(regression from edd47ed5)

### Reference issue
Ref #56

### DIFF
--- a/abssendtimeextension.go
+++ b/abssendtimeextension.go
@@ -5,26 +5,18 @@ import (
 )
 
 const (
-	absSendTimeExtensionSize = 4
+	absSendTimeExtensionSize = 3
 )
 
 // AbsSendTimeExtension is a extension payload format in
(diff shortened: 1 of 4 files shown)


sgotti:
@at-wat I was just thinking if, after #56, the various extensions should be updated to just parse the extension payload and not the header. This is probably a breaking API change but also #56 introduced  a breaking API change.

at-wat:
> This is probably a breaking API change but also #56 introduced a breaking API change.

Yes, this changes API again, but I think this change had to be a part of #56.
I'm using AbsSendTime extension in the packetizer in my project. The current master branch can't handle it, even after migrating the API change.

at-wat:
I'll add a tag `v1.5.3` as a bug fix of v1.5.

PR #67, opened 2020-05-19
Fix TransportLayerCC

adwpc:
Fix TransportLayerCC

#### Description
Fix ID and len

#### Reference issue
Fixes #...

### DIFF
--- a/transportccextension.go
+++ b/transportccextension.go
@@ -6,7 +6,8 @@ import (
 )
 
 const (
-	transportCCExtensionSize = 4
+	// transport-wide sequence
+	transportCCExtensionSize = 2
 )
 
 var (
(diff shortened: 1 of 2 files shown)


transportccextension.go
```diff
@@ -6,7 +6,8 @@ import (
 )
 
 const (
-	transportCCExtensionSize = 4
+	// transport-wide sequence and zero padding
+	transportCCExtensionSize = 3
```
at-wat, line 10: Padding is automatically added by `Marshal()` now.
https://github.com/pion/rtp/blob/e9659f71a59bf9dd22935d98e29fead23428547e/packet.go#L313-L317

adwpc: Fixed.

PR #74, opened 2020-07-12
Support VP9 Scalability Structure (SS)

hakobera:
#### Description

Add support for parsing VP9 scalability structure to avoid error
when a packet contains scalability structure (V=1).

#### Reference issue
None


Sean-Der:
The PR looks great to me! Thank you so much for the contribution.

Great to see you here @hakobera if you are interested I would love to have you more involved in pion. We need all the help we can get :)

PR #106, opened 2020-12-12
Fix Unmarshal to used struct

at-wat:
Extensions field was not cleared on Unmarshal and data was appended
to the old extension slice.
This fixes Extensions field to have correct number of data even if
the slice has some data before calling Unmarshal.

Reuse Extensions and CSRC fields.

### master branch:
```
BenchmarkUnmarshal/SharedStruct-32      6965502    158 ns/op    370 B/op    1 allocs/op
**^Unmarshal result was wrong**
BenchmarkUnmarshal/NewStruct-32         9876006    121 ns/op    104 B/op    3 allocs/op
```
### this:
```
BenchmarkUnmarshal/SharedStruct-32     61906088   19.1 ns/op      0 B/op    0 allocs/op
BenchmarkUnmarshal/NewStruct-32         9698794    123 ns/op    104 B/op    3 allocs/op
```


Sean-Der (approved):

Sean-Der:
@at-wat tagged also!

Issue #107, opened 2020-12-15
Support AVC depacketizer


debiandebiandebian:
## Summary

The current implementation enforces `Annex B` packaging. If each `NALU` should be picked up for
- repacking it into `AVC` for `Matroska` container, or
- checking if all key frame elements (i.e., `SPS` + `PPS` + `IDR`) are present in an aggregate `RTP` packet,

another iteration based on start code should be done. `AVC` packaging is inherently good for locating individual elements, and giving another payloader does not break anything.

## Motivation

Same as above.

## Describe alternatives you've considered

It would be best to return a slice of `NALU`s if possible.

Sean-Der:
We could give `H264Payloader` a `IsAVC bool` key, and then the Payloader could take different paths.

I think this would be a pretty easy change! Thanks for requesting @debiandebiandebian if you are interested in contributing happy to help

debiandebiandebian:
The title was misleading. I meant `H264Packet`, not `H264Payloader`. So, I was talking about `RTP` depacketizer. But, I suppose you catch it right.

Come to think of it, it would be difficult to implement. For example, the first packet of fragmentation units does not tell anything about the entire `NALU` length.

Sean-Der:
Oh yea, I support that as well! Since this isn't an API breaking change I would love to take it

debiandebiandebian:
The reason why I thought it's difficult to implement was
- the entire `NALU` length is unknown until the final [`FUs`](https://tools.ietf.org/html/rfc6184#section-5.8) packet is received, and
- `samplesbuilder` spits out the result immediately for each `RTP` input. 

It was unclear to me how we go back to the first result, and re-write the length portion already given to the user.

But, this trick looks good to me:
- `return []byte{}` for non-final `FUs` packets, otherwise
- aggregate & return the result with preceding `NALU` length.

debiandebiandebian:
If you're fine with the direction above, I would love to give it a shot.

Sean-Der:
Yes! Please make a PR, happy to help however I can :)

debiandebiandebian:
@Sean-Der Oops, I didn't think you would merge it right away.

I want to point out one important pitfall from this PR. Any existing application depending on this module would break if
- a single `H264Packet` instance is accessed from multiple goroutines.

Back then, we had no instance fields, so users might have used a single instance from all over the places. In this case, `fuaBuffer` will be stuffed with two or more different `RTP` stream packets.

[...]

PR #110, opened 2020-12-19
Support AVC depacketizer

debiandebiandebian:
#### Description
- Add `IsAVC` flag to choose between `Annex B` or `AVC` packaging,
- Add `fuaBuffer` to keep non-final `FU-A` packets until we know the complete `NALU` length, and
- Add `fuaEndBitmask` to determine if we have to flush `FU-A` buffer.

#### Reference issue
https://github.com/pion/rtp/issues/107


debiandebiandebian:
@Sean-Der Can you review this PR?

Sean-Der (approved):

Sean-Der:
Fantastic work! I am so sorry this took so long @debiandebiandebian :( 

I also added you to the pion organization. The holiday + going back to work has really took up my time. This is a fantastic PR, my slowness had nothing to do with this PR.

thanks

robin-raymond:
@debiandebiandebian There is an issue with this code where is will break pion/webrtc h264writer_test.go

The reason is because the writer uses:
````
	data, err := (&codecs.H264Packet{}).Unmarshal(packet.Payload)
````

Basically, it creates a H264Packet object on the fly to unmarshal to... but the unit test first writes the h264 FUA START, then an H264 FU END packet, but the first packet unmarshal returns data array with 0 content, the second only returns the END data but not the contents in fuaBuffer because it's been discarded between the previous write.

I'm uncertain the expected fix. I think the H264 packet object be reused in the webrtc h264 writer until the close is called to ensure any buffering in the fuaBuffer is preserved, but I worry about the side effects.

cc: @Sean-Der

PR #112, opened 2021-01-08
Pack SPS and PPS in one STAP-A packet

baiyufei:
Chrome need SPS and PPS in one STAP-A, and timestamp of STAP-A should be same with next I-Frame. Otherwise Chrome will discard SPS and PPS before receiving I frame.

#### Reference issue
Fixes https://github.com/pion/webrtc/issues/1625

### DIFF
--- a/codecs/h264_packet.go
+++ b/codecs/h264_packet.go
@@ -6,12 +6,18 @@ import (
 )
 
 // H264Payloader payloads H264 packets
-type H264Payloader struct{}
+type H264Payloader struct {
+	spsNalu, ppsNalu []byte
+}
 
 const (
-	stapaNALUType = 24
-	fuaNALUType   = 28
-	fubNALUType   = 29
+	stapaNALUType  = 24
+	fuaNALUType    = 28
+	fubNALUType    = 29
+	spsNALUType    = 7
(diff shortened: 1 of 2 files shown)


caiiiycuk:
Hi @baiyufei,  you fixes work fine with chrome. But FF shows error and does not start streaming.
Error:
```
Answer tried to set recv when offer did not set send
```

It also reproduced in your test repo.
--
Firefox 89.0.2 Ubuntu

Sean-Der:
This is great @baiyufei thank you so much! Sorry for the delay, I am finally getting around to a `v3.1.0`

What do you think of emitting the STAP-A [here](https://github.com/pion/rtp/pull/112/files#diff-5eca440e933baa4c5491dcd9ccc7c1f4d807994e6c5b54c304228c34fcc3c963R92) instead? Do we need to wait for the next call to packetize?

Sean-Der:
@baiyufei I just made the change, if you agree I will merge that. If it needs to stay as it was before I will rollback.

thanks!

Sean-Der (approved):

Sean-Der:
Confirmed your fix is needed as is, nice work @baiyufei !

PR #115, opened 2021-01-28
Robin 20210119 seqnum

robin-raymond:
#### Reference issue
Fixes #...

### DIFF
--- a/packetizer.go
+++ b/packetizer.go
@@ -11,8 +11,9 @@ type Payloader interface {
 
 // Packetizer packetizes a payload
 type Packetizer interface {
-	Packetize(payload []byte, samples uint32, skippedSamples uint32) []*Packet
+	Packetize(payload []byte, samples uint32) []*Packet
 	EnableAbsSendTime(value int)
+	SkipSamples(skippedSamples uint32)
 }
 
 type packetizer struct {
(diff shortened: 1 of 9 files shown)


Sean-Der:
Nice! This is fantastic, didn't think about expressing discontinuity at all before. This is great.

The API break is a tough pill to swallow, I don't think we can do that in `/v3` of Pion WebRTC. What do you think of expressing discontunity by returning a `Sample{Duration: time.Second, Data: []}`?

* Not an API break
* SampleBuilder can emit gaps that way and a user can see they happened!

robin-raymond:
Sure, that would work too. I can fix it as appropriate. But I think that code is on the webrtc repo not the rtp repo?

packetizer.go
```diff
@@ -11,7 +11,7 @@ type Payloader interface {
 
 // Packetizer packetizes a payload
 type Packetizer interface {
-	Packetize(payload []byte, samples uint32) []*Packet
+	Packetize(payload []byte, samples uint32, skippedSamples uint32) []*Packet
```
tarrencev, line 14: Can we land this change without this breaking change?

Sean-Der:
@rob-deutsch @tarrencev sorry for the initial pushback, I say we go forward with these changes!

I will fix the merge conflict myself and then we can review one more time and get in :) we have quite a few breaking RTP changes that I have been sitting on. This would be a great time to do it.

Issue #90, opened 2020-09-23
Remove Header.PayloadOffset and Packet.Raw


Sean-Der:
We have duplication of data `Raw`/`PayloadOffset` and `Payload`.

We need to drop these fields so we have a less fragile API

ffmiyo:
@Sean-Der  Is `Header.PayloadOffset` a must go? It is needed during unmarshal to split between header and payload.  
An alternative is to return as Int from `Header.Unmarshal()`.

Issue #79, opened 2020-09-03
CSRC Unmarshal / Marshal diff


Antonito:
### Your environment.
  - Version: v1.6.0
  - Browser: N/A

### What did you do?
```go
package rtp_test

import (
	"bytes"
	"testing"

	"github.com/stretchr/testify/assert"

	"github.com/pion/rtp"
	"github.com/pion/rtp/codecs"
)

func Test_RTP_Marshal_Unmarshal(t *testing.T) {
	multiplepayload := make([]byte, 128)
	packetizer := rtp.NewPacketizer(100, 98, 0x1234ABCD, &codecs.G722Payloader{}, rtp.NewRandomSequencer(), 90000)
	packets := packetizer.Packetize(multiplepayload, 1000)

	rawPkts := make([][]byte, 0, 1400)
	for _, pkt := range packets {
		raw, err := pkt.Marshal()
		assert.Nil(t, err)

		rawPkts = append(rawPkts, raw)
	}

	for ndx, raw := range rawPkts {
		expectedPkt := packets[ndx]
		pkt := &rtp.Packet{}

		err := pkt.Unmarshal(raw)
		assert.Nil(t, err)

		assert.EqualValues(t, len(raw), pkt.MarshalSize())
		assert.EqualValues(t, expectedPkt.MarshalSize(), pkt.MarshalSize())

		assert.EqualValues(t, expectedPkt.Version, pkt.Version)
		assert.EqualValues(t, expectedPkt.Padding, pkt.Padding)
		assert.EqualValues(t, expectedPkt.Extension, pkt.Extension)
		assert.EqualValues(t, expectedPkt.Marker, pkt.Marker)
		assert.EqualValues(t, expectedPkt.PayloadOffset, pkt.PayloadOffset)
		assert.EqualValues(t, expectedPkt.PayloadType, pkt.PayloadType)
		assert.EqualValues(t, expectedPkt.SequenceNumber, pkt.SequenceNumber)
		assert.EqualValues(t, expectedPkt.Timestamp, pkt.Timestamp)
		assert.EqualValues(t, expectedPkt.SSRC, pkt.SSRC)
		assert.EqualValues(t, expectedPkt.CSRC, pkt.CSRC)
		assert.EqualValues(t, expectedPkt.ExtensionProfile, pkt.ExtensionProfile)
		assert.EqualValues(t, expectedPkt.Extensions, pkt.Extensions)
		assert.EqualValues(t, expectedPkt.Payload, pkt.Payload)

[...]

PR #118, opened 2021-02-03
Implement packet Clone method

### LINKED ISSUE #88: Add Clone to rtp.Packet
In my SFU, packets go through a number of data structures, and finally end up in a static buffer that is passed to `Track.WriteRTP`.  This works fine with ordinary tracks, but fails when trying to write data to disk: `SampleBuilder.Push` requires the packet to never be overwritten.  This does not appear to be documented.  (My personal opinion is that `Push` should clone the packet, but as a second choice I'd be okay if the requirement that the `Packet` and its underlying buffer never be reusied were clearly documented.)

What is more, there is no easy way to clone a packet.  I'm currently using the following code:

    func clonePacket(packet *rtp.Packet) *rtp.Packet {
    	buf, err := packet.Marshal()
    	if err != nil {
    		return nil
    	}
    	var p rtp.Packet
    	err = p.Unmarshal(buf)
    	if err != nil {
    		return nil
    	}
    	return &p
    }

My suggestions are therefore:

 * document that `SampleBuilder.Push` requires its parameter to never be modified;
 * provide a function `Packet.Clone()` that clones a packet efficiently;
 * in v3, avoid the complication by having `SampleBuilder.Push` copy the packet internally.

ffmiyo:
Clone performs deep copy on a packet to produce equivalent but independently mutable packet.

Fixes #88

### DIFF
--- a/packet_test.go
+++ b/packet_test.go
@@ -1138,6 +1138,59 @@ func TestRoundtrip(t *testing.T) {
 	}
 }
 
+func TestCloneHeader(t *testing.T) {
+	h := Header{
+		Marker:           true,
+		Extension:        true,
+		ExtensionProfile: 1,
+		Extensions: []Extension{
+			{0, []byte{
+				0xFF, 0xFF, 0xFF, 0xFF,
+			}},
+		},
+		Version:        2,
+		PayloadType:    96,
+		SequenceNumber: 27023,
+		Timestamp:      3653407706,
+		SSRC:           476325762,
+		CSRC:           []uint32{},
+	}
+	clone := h.Clone()
+	if !reflect.DeepEqual(h, clone) {
+		t.Errorf("Cloned clone does not match the original")
+	}
+
+	h.CSRC = append(h.CSRC, 1)
+	if len(clone.CSRC) == len(h.CSRC) {
+		t.Errorf("Expected CSRC to be unchanged")
+	}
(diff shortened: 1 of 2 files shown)


Sean-Der:
Woah this is awesome @ffmiyo! 

This might also be a good chance to do a `/v2` break. Would you be interested? We can get it into `pion/webrtc` right away as well. While it would be a breaking change, I don't see anyone that depends on it.

* https://github.com/pion/rtp/issues/90
* https://github.com/pion/rtp/issues/79

ffmiyo:
@Sean-Der Ah right, should've start with # 90. Sure, I'll try to do this.

packet_test.go
```diff
@@ -1138,6 +1138,60 @@ func TestRoundtrip(t *testing.T) {
...
+		Marker:           true,
+		Extension:        true,
+		ExtensionProfile: 1,
+		Extensions: []Extension{
+			{0, []byte{
+				0xFF, 0xFF, 0xFF, 0xFF,
+			}},
+		},
+		Version:        2,
+		PayloadOffset:  20,
```
aler9, line 1152: must be updated to account for #119 by same author

aler9 (changes requested):
must be updated to account for #119 by same author

Sean-Der (approved):

PR #123, opened 2021-04-19
Packetization of v8 w/ PictureID and skip samples

robin-raymond:
Packetization of v8 w/ PictureID and skip samples
    
VP8 should include the picture ID for the sake of SLI when packetizing. The number of skipped samples during packetization needs to be factored as well.


robin-raymond:
will replace https://github.com/pion/rtp/pull/115

PR #126, opened 2021-04-21
Packetization of v8 w/ PictureID and skip samples

robin-raymond:
Packetization of v8 w/ PictureID and skip samples

VP8 should include the picture ID for the sake of SLI when
packetizing. The number of skipped samples during packetization
needs to be factored as well.


Sean-Der (approved):

at-wat:
@robin-raymond In my understanding of RFC7741, PictureID is not mandatory and required only if SLI is used.
Is there a specification that PictureID must be always added? Otherwise, I think the feature should be able to be opt-out.

PR #132, opened 2021-05-22
Drop /v2

### LINKED ISSUE #129: Please revert switch to v2
Commit 863e8351e01da61813308c8689c2aea684402450 updates the package version to v2.  This causes a lot of churn for no good reason, since commit 6b516fba426bbd973d1fef43de8878156056160e merely removes some fields that nobody is using anyhow (that's why they're being removed).

An alternative would be to revert both commits.

If reverting this commit is not possible, then we need an update to pion/webrtc ASAP, since the samplebuilder takes a v1 packet.

aler9:
Since /v2 is causing interaction problems with other parts of the
Pion project, i'm removing my previous commit. The issue must
anyway dealt with in a future discussion with the community.

Fixes #129

#### Reference issue
Fixes #...

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/pion/rtp/v2
+module github.com/pion/rtp
 
 go 1.13
 
--- a/packetizer_test.go
+++ b/packetizer_test.go
@@ -6,7 +6,7 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


Sean-Der:
Hey @aler9 

Sorry to leave you hanging on this. IMO we just need to move forward with /v2 @at-wat opened PRs upgrading stuff across the project so I think we are good :)

PR #140, opened 2021-07-07
Remove the IsDetectedFinalPacket method from Depacketizer

### LINKED ISSUE #139: Interface change breaks existing code
Revision #34c6ce978ad576469ea43860fecc48fad612c7c8 added a new method to the Depacketizer interface, thus breaking all existing code that implements a Depacketizer.

The method itself was introduced in #9e081b57389c3b1d859971912cb7e8e3d194d1cc, and is basically a verbose and error-prone way of saying
```
if codec is video {
     return packet.Mark
}
return true
```
I therefore recommend that both changes above should be reverted.  If the functionality is needed (which I doubt), it should be implemented by checking whether the codec is doing video, and doing the right thing without imposing an extra method on all depacketisers, which is difficult to check and causes massive code duplication.

jech:
Not only does this break all code that attempts to implement
a depacketizer, it is just a very verbose and error-prone way
of saying

  if codec is video {
      return packet.Mark
  }
  return true

It is better to check the codec type at the calling site rather
than imposing a new method on all depacketisers, which causes
code duplication and is very difficult to check.

Fixes #139.

### DIFF
--- a/codecs/h264_packet.go
+++ b/codecs/h264_packet.go
@@ -165,12 +165,6 @@ func (p *H264Packet) doPackaging(nalu []byte) []byte {
 	return append(annexbNALUStartCode(), nalu...)
 }
 
-// IsDetectedFinalPacketInSequence returns true of the packet passed in has the
-// marker bit set indicated the end of a packet sequence
-func (p *H264Packet) IsDetectedFinalPacketInSequence(rtpPacketMarketBit bool) bool {
-	return rtpPacketMarketBit
-}
-
 // Unmarshal parses the passed byte slice and stores the result in the H264Packet this method is called upon
 func (p *H264Packet) Unmarshal(payload []byte) ([]byte, error) {
 	if payload == nil {
--- a/codecs/opus_packet.go
+++ b/codecs/opus_packet.go
@@ -19,12 +19,6 @@ type OpusPacket struct {
 	Payload []byte
 }
 
-// IsDetectedFinalPacketInSequence returns true as all opus packets are always
-// final in a sequence
-func (p *OpusPacket) IsDetectedFinalPacketInSequence(rtpPacketMarketBit bool) bool {
-	return true
-}
-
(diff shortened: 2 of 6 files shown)


jech:
CC @robin-raymond

jech:
Please see https://github.com/jech/samplebuilder which implements the exact same functionality without forcing an extra method for all implementers of depacketizer.

jech:
See https://github.com/pion/webrtc/pull/1882 for a clean implementation of the same functionality.

Sean-Der:
@jech One downside of this is that this requires users to configure the `SampleBuilder`. Do you see any benefits of configuration being bundled with the codec?

jech:
There are two possible choices:
  1. partition boundaries are the responsibility of the codec;
  2. partition boundaries are the responsibility of the caller.
Right now, we are inconsistent: partition heads are the responsibility of the caller (WithPartitionHeadChecker), while partition tails are the responsibility of the codec.

Pull request https://github.com/pion/webrtc/pull/1882 makes things consistent again by moving parition tail checking into the caller (case (2) above).  This has the advantage of a consistent API and of only using a small number of well-defined interfaces.  The disadvantage is that the caller needs to pass the right partition head and tail checkers to each sample builder.

If there is a strong consensus in favour of making things automatic (which I dislike, too much magic makes code undebuggable), then I can try to think about how to do that cleanly, without adding a new method to each and every new codec (C++ style).

Sean-Der:
I am in favor of making things automatic (having an interface)

* Most users will not be implementing their own codecs. I don't want to require boilerplate code for the SampleBuilder to work for them.
* For developers who wish to customize a codec they can embed and provide their own methods
* For developers who wish to implement their own codecs entirely I can see the frustration. 

I took a look through github at the users of SampleBuilder and that felt to line up with most users. I don't have much data (or user interviews I can pull from). If this blocks you from implementing anything @jech I am happy to re-visit. 

I am sorry that we provide the interface and didn't bump the API. It is unfortunate, will try to set up tooling so it doesn't happen again.

PR #144, opened 2021-07-14
Revert the polarity of EnablePictureID

jech:
Replace the EnablePictureID flag by OmitPictureID, which has the
opposite polarity.  This means that the zero value enables the
PictureID, which is the better default in most cases.

Related to https://github.com/pion/webrtc/pull/1875

### DIFF
--- a/codecs/vp8_packet.go
+++ b/codecs/vp8_packet.go
@@ -2,8 +2,8 @@ package codecs
 
 // VP8Payloader payloads VP8 packets
 type VP8Payloader struct {
-	EnablePictureID bool
-	pictureID       uint16
+	OmitPictureID bool
+	pictureID     uint16
 }
 
 const (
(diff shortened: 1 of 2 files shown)


Sean-Der:
I am not familiar with VP8 RTP Packetization at all, @at-wat mind taking a look?

at-wat:
As RTP is not only for WebRTC, I believe optional things in RFC should be optional in the implementation.
And it's better to keep polarity of the options consistent. It wouldn't be good if `Disable*` and `Enable*` are mixed a lot.

GStreamer rtpvp8pay also sets default `picture-id-mode` to `none`.
https://gstreamer.freedesktop.org/documentation/rtp/rtpvp8pay.html?gi-language=c#rtpvp8pay:picture-id-mode

at-wat (commented):
As RTP is not only for WebRTC, I believe optional things in RFC should be optional in the implementation.
And it's better to keep polarity of the options consistent. It wouldn't be good if `Disable*` and `Enable*` are mixed a lot.

GStreamer rtpvp8pay also sets default `picture-id-mode` to `none`.
https://gstreamer.freedesktop.org/documentation/rtp/rtpvp8pay.html?gi-language=c#rtpvp8pay:picture-id-mode

Sean-Der:
Thanks for the quick response @at-wat! Makes sense to me

PR #155, opened 2021-11-22
Add PaddingSize to rtp.Packet

boks1971:
When a packet has padding, the padding size information
is lost during unmarshal. Add a PaddingSize field to
Packet structure.

Also support marshalling on an RTP packet with padding.

Testing:
--------
Modified/added unit tests.

### DIFF
--- a/packet.go
+++ b/packet.go
@@ -30,7 +30,8 @@ type Header struct {
 // Packet represents an RTP Packet
 type Packet struct {
 	Header
-	Payload []byte
+	Payload     []byte
+	PaddingSize byte
 }
 
 const (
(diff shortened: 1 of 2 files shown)


Sean-Der (approved):

bixycler:
> Also support marshalling on an RTP packet with padding.

But even with marshalled packet, the _plaintext will always be unmarshalled and its **padding info is thrown away**_: both `webrtc.TrackLocalStaticRTP.Write(b []byte)` and `srtp.SessionSRTP.write(b []byte)` just `p.Unmarshal(b)` then call the "writeRTP()" counterparts, and the packet is not passed down as a whole but only its **header** and **payload** (_missing **PaddingSize**_) are passed down via `b.writeStream.WriteRTP(&p.Header, p.Payload)`, to `TrackLocalWriter`,... and to `srtp.Context.encryptRTP(dst []byte, header *rtp.Header, payload []byte)`. See pion/webrtc#2403

PR #162, opened 2022-01-03
Adds zero-copy header extensions

kevmo314:
This change adds an interface and impl for zero-copy extensions. The code is forked from the existing `packet.go` but those files will be migrated later.


davidzhao (approved):
great work Kevin!

PR #172, opened 2022-04-07
Start AV1 implementation

Sean-Der:
WIP


Sean-Der:
Here is an example of a IVFWriter that is when passed AV1 RTP packets saves to disk successfully!

Sean-Der:
@enobufs I also need to think/design how the OBU Cache works and take sequence numbers into account. 

We should associate the cache with a sequence number. If we have gaps we should just evict the cache. It is a harder problem, but I am sure we can come up with a good design.

codecs/av1_packet.go
```diff
@@ -0,0 +1,138 @@
...
+		if p.obuBuffer == nil {
+			return
+		}
+		obuElement = append(p.obuBuffer, obuElement...)
+		p.obuBuffer = nil
+	}
+	p.OBUs = append(p.OBUs, obuElement)
+}
+
+// Unmarshal parses the passed byte slice and stores the result in the VP8Packet this method is called upon
```
boks1971, line 82: VP8 typo?

boks1971 (approved):
Awesome!

enobufs:
> We should associate the cache with a sequence number.

@Sean-Der I think so too because packet loss could occur over the network and NACK may not be able restore them or applicable, but this is done outside the av1_packet.go.  We could update the IVFWriter example to tolerate the case of the discontinuity.

codecs/av1_packet.go
```diff
@@ -0,0 +1,138 @@
...
+		obuElementLength, bytesRead uint
+		err                         error
+	)
+	for i := 1; ; i++ {
+		if currentIndex == uint(len(payload)) {
+			break
+		}
+
+		// If W bit is set the last OBU Element will have no length header
+		if byte(i) == p.W {
```
enobufs, line 113: I remember W can't be greater than 3 and should fail.

codecs/av1_packet.go
```diff
@@ -0,0 +1,138 @@
...
+	W byte
+
+	// N: MUST be set to 1 if the packet is the first packet of a coded video sequence, and MUST be set to 0 otherwise.
+	N byte
+
+	OBUs [][]byte
+
+	// Buffer for fragmented OBU. If Unmarshal is called on a RTP Packet
+	// that doesn't contain a fully formed OBU
+	obuBuffer []byte
```
enobufs, line 66: I'm not sure if we should assume that the same instance of AV1packet can be reused for the subsequent RTP packets.  Users can see Z and Y flags. We can simply leave the fragment in the first and the last OBUs array IMO.

enobufs (commented):
I left a couple of my comments. Let me know what you think.

codecs/av1_packet.go
enobufs: If that you think is ugly, we could ditch the packet-by-packet "Unmarshal" semantics and use a callback style, e.g. OnOBUElement(func([]byte) error)... I may like it better IMHO.

codecs/av1_packet.go
Sean-Der: Fixed!

codecs/av1_packet.go
Sean-Der: Added `errWGreaterThenThree`

codecs/av1_packet.go
Sean-Der: @enobufs So one nice thing about keeping `Unmarshal` is that it satisfies the `rtp.Depacketizer` interface. If I knew everything I know now I would have designed it differently :/

We assume that users keep the same H264 Packet across Unmarshals and has seemed to work decently well. I think we should make it more robust though. Checking things like SequenceNumbers etc...

For a new version of `pion/rtp` it would be great to completely rethinking how unmarshaling video works. Maybe something like `OnFrame` so it could work for all codecs?

kevmo314: I'm rather against the callback style compared to the current unmarshal pattern. imo the callback style is not very go-thonic.

Re: rethinking how unmarshal works, @Sean-Der I have done a ReadFrame() pattern that pulls frames before. Basically you can pass in a byte reader and it returns a frame reader so something like:

```
type RTPReader interface {
  ReadRTP() (*rtp.Packet, error)
}

type FrameReader interface {
  ReadFrame() (*Frame, error)
}
```

then the unmarshaller:

```
DepacketizeH264(RTPReader) FrameReader
```

This allows on-demand-style unmarshalling, matches stdlib, and largely the motivation behind the rtpio package was to facilitate this sort of pattern for higher level objects.

I'm not against requiring the reuse of a packet type though (ie keep the current pattern). I think it's not that unreasonable.

kevmo314 (approved):
Looks good to me, super exciting!

codecs/av1_packet.go
Sean-Der: I love that! What if we added a `AV1Frame` `VP9Frame` with Unmarshal methods @enobufs @kevmo314 ?

then we can get both RTP payload awareness and higher level ‘I just want a video frame’ API

Sean-Der: Will update the PR with that today

kevmo314: Sounds good to me! Yeah if we go through that route (and assuming all the codecs do), we could actually remove `SampleBuilder` completely, since all the codecs would be able to provide sample construction through an API.

And for some wishful thinking we can even pipe this all the way to `io.Writer` so depacketizing and writing to a file could be as simple as:

```
go io.Copy(f, rtp.NewIVFWriter(rtp.ParseAV1(av1track)))
```

😄

Sean-Der: @kevmo314 @enobufs What do you think of that?

I also see us keeping around the SampleBuilder? In the future it will only be concerned with filling holes/handling jitter? Since that code is codec agnostic we don't want to put it in a specific RTP codec.

enobufs: The read frame semantics is good and I agree it's better than callback style.
I realize I did not communicate my point very well. It was about `obuBuffer` being inside AV1Packet.

I can see two layers in this context (high to low):
* OBU Frame construction (Inside the IVFWriter example)
* RTP Payload parsing for AV1 (AV1Packet)

I believe the `obuBuffer` belongs to the "OBU Frame construction".

When I was trying to add the following code to IVFWrite to handle cases for missing RTP packets, it became clear that there needs to be a way to invalidate `obuBuffer` but it's not possible because it is a private member. I feel obuBuffer does not belong there. If I understood the above plan correctly, it should be in the FrameReader IMO.

AV1 support by Pion is super exciting!!

codecs/av1_packet.go
```diff
@@ -0,0 +1,116 @@
...
+// +-+-+-+-+-+-+-+-+
+// |Z|Y| W |N|-|-|-|
+// +-+-+-+-+-+-+-+-+
+//
+// https://aomediacodec.github.io/av1-rtp-spec/#44-av1-aggregation-header
+type AV1Packet struct {
+	// Z: MUST be set to 1 if the first OBU element is an
+	//    OBU fragment that is a continuation of an OBU fragment
+	//    from the previous packet, and MUST be set to 0 otherwise.
+	Z byte
```
at-wat, line 41: I think single bit field is better to be `bool` like
https://github.com/pion/rtp/blob/b08333e6c3a2db0ec7c388c81541d91d78a9ae70/codecs/vp9_packet.go#L115-L153

Sean-Der: Fixed!

at-wat:
@Sean-Der I'm not very sure `type AV1Frame struct` belongs to `pion/rtp/codecs` since it was mainly for RTP packet definitions for each codec.
It might be better to move it under `pion/rtp/pkg/`

Sean-Der:
Sorry I thought you were done reviewing @at-wat :( 

That works for me! `pion/rtp/pkg/frame.AV1` maybe? I am not at my computer (heading to bed). I will move in the morning if you agree! Also feel free to move yourself and use admin override or I can approve+merge in the morning.

enobufs:
@Sean-Der should I file a new issue? https://github.com/pion/rtp/pull/172#discussion_r846829734

Sean-Der:
@enobufs I can fix it all now!

I did move the [obuBuffer](https://github.com/pion/rtp/blob/master/codecs/av1_frame.go#L10) to the AV1Frame in the PR. Any other changes I missed?

PR #175, opened 2022-04-11
Move AV1 Frame to new frame pkg

Sean-Der:
Codecs only contains RTP Payload processing


codecs/av1_packet.go
```diff
@@ -1,7 +1,7 @@
 package codecs
 
 import (
-	"github.com/pion/rtp/v2/pkg/obu"
+	"github.com/pion/rtp/pkg/obu"
```
at-wat, line 4: In my understanding, `master` branch is v2 and `v1` branch is v1.
Is it intended to use v1/pkg/obu from v2/codecs?

at-wat (approved):
https://github.com/pion/rtp/commit/9d538e66480c03d08a07beb8b845c7b25c80130a looks good to me

PR #192, opened 2022-06-24
Advance Timestamp when empty payload is packetized

adamroach:
#### Description

Currently, the `Packetize` method has a sanity check that the payload
being packetized is not of zero length; if is, then the function returns
without doing anything. This interacts poorly with pion's binding to
OpenH264, which simply returns payloads of zero length when `encode()`
is called but a frame must be skipped for bandwidth limitation reasons.

The most straightforward use of these two APIs with each other leads to
a situation where an application reads an empty payload from OpenH264
and writes it into the Payloader directly. When this happens, the count
of samples is not advanced, leading to incorrect timestamps in the
generated RTP packets (which can cause A/V sync issues).  This change
simply ensures that the Packetizer Timestamp gets updated under such
circumstances.


Sean-Der:
Added a little test, merging now :)

PR #196, opened 2022-07-29
Add AbsCaptureTimeExtension and PlayoutDelayExtension

kevmo314:
Both of these are experimental RTP header extensions defined in
libwebrtc.

### DIFF
--- a/abscapturetimeextension.go
+++ b/abscapturetimeextension.go
@@ -1,0 +1,87 @@
+package rtp
+
+import (
+	"encoding/binary"
+	"time"
+)
+
+const (
+	absCaptureTimeExtensionSize         = 8
+	absCaptureTimeExtendedExtensionSize = 16
+)
+
+// AbsCaptureTimeExtension is a extension payload format in
+// http://www.webrtc.org/experiments/rtp-hdrext/abs-capture-time
+// 0                   1                   2                   3
+// 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |  ID   | len=7 |     absolute capture timestamp (bit 0-23)     |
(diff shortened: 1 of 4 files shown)


abscapturetimeextension.go
```diff
@@ -0,0 +1,90 @@
...
+
+// EstimatedCaptureClockOffsetDuration produces the estimated time.Duration represented by this extension.
+func (t AbsCaptureTimeExtension) EstimatedCaptureClockOffsetDuration() *time.Duration {
+	if t.EstimatedCaptureClockOffset == nil {
+		return nil
+	}
+	offset := *t.EstimatedCaptureClockOffset
+	duration := time.Duration((offset>>32)&0x7FFFFFFF)*time.Second + time.Duration((offset&0xFFFFFFFF)*1e9/(1<<32))*time.Nanosecond
+	if offset < 0 {
+		duration *= -1
```
mengelbart, line 68: Why absolute and not signed as transmitted?

playoutdelayextension.go
```diff
@@ -0,0 +1,47 @@
...
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |  ID   | len=2 |       MIN delay       |       MAX delay       |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+type PlayoutDelayExtension struct {
+	minDelay, maxDelay uint16
+}
+
+// Marshal serializes the members to buffer
+func (p PlayoutDelayExtension) Marshal() ([]byte, error) {
+	if p.minDelay >= playoutDelayMaxValue || p.maxDelay >= playoutDelayMaxValue {
```
mengelbart, line 28: I think `playoutDelayMaxValue` itself should be allowed, i.e. this line should use `>` instead of `>=`?

kevmo314: Done, thanks for catching that.

abscapturetimeextension.go
kevmo314: The prior version was signed as transmitted, just the way I was doing it first stripped the sign. I updated the code to be more clear.

mengelbart (approved):
Thanks for the updates!

PR #197, opened 2022-08-11
Exports MinDelay and MaxDelay

kevmo314:
The header cannot be used outside of this package currently.


Sean-Der:
Was done cd52e6b5

PR #200, opened 2022-11-03
Add H265SafariPayloader

erivni:
#### Description
Added H265SafariPayloader.
based on https://github.com/AlexxIT/Blog/issues/5

@kevmo314


kevmo314:
Interesting, really weird that Safari decided to create their own payload format. This has some rather odd implications for interoperability, however. Suddenly the payloader needs to know about where the stream is sent :(

Is the Safari payloader compatible with ffmpeg RTP and others? If so, maybe it would be better to adopt the Safari format as "standard"...

Change looks good to me though, thanks for figuring this out :)

kevmo314 (approved):

erivni:
yes, we don't know why safari decided to use their own payload format.
vlc, ffplay and gst doesn't support this format, so i believe the standard should remain as is.

i changed the code to use 'copy' instead of 'append' for memory efficiency.

@shlompy, thank you as well!

PR #203, opened 2023-01-06
Fix parsing of VP8 packets with degenerate header

jech:
Fix parsing of VP8 packets with degenerate header

All of the fields in the VP8 header except the first byte are
optional.  We used to reject VP8 packets smaller than 4 bytes,
which is incorrect.

There are two cases where such packets may appear on the wire.
GStreamer's WebRTC implementation generates VP8 streams with
no picture id, and one-byte headers.  It will occasionally
generate packets that are below 4 bytes, and which we used
to reject.

The second use case is more theoretical.  According to RFC 7741
Section 4.4, a packetizer may ignore VP8 partition boundaries.
If it splits a packet outside of a partition boundary, it may
generate a packet with S=0 and a one-byte header.

### DIFF
--- a/codecs/vp8_packet.go
+++ b/codecs/vp8_packet.go
@@ -127,12 +127,11 @@ func (p *VP8Packet) Unmarshal(payload []byte) ([]byte, error) {
 
 	payloadLen := len(payload)
 
-	if payloadLen < 4 {
-		return nil, errShortPacket
-	}
-
 	payloadIndex := 0
 
+	if payloadIndex >= payloadLen {
+		return nil, errShortPacket
+	}
 	p.X = (payload[payloadIndex] & 0x80) >> 7
 	p.N = (payload[payloadIndex] & 0x20) >> 5
 	p.S = (payload[payloadIndex] & 0x10) >> 4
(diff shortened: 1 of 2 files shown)


codecs/vp8_packet.go
```diff
@@ -127,12 +127,11 @@ func (p *VP8Packet) Unmarshal(payload []byte) ([]byte, error) {
 
 	payloadLen := len(payload)
 
-	if payloadLen < 4 {
-		return nil, errShortPacket
-	}
-
 	payloadIndex := 0
 
+	if payloadIndex >= payloadLen {
```
tmatth, line 132: Just a readability thing here, given that `payloadIndex` has just been set to 0, IMO this would be easier to grok as
`if payloadLen == 0` as this is the only case you're really guarding against (i.e. `payloadIndex` can't be greater than `payloadLen` so that leaves only the equality case, and they'd only be equal if `payloadLen` was 0).

davidzhao (approved):
lgtm. agree with @tmatth's comment re: readability. would keep it slightly more readable.

PR #205, opened 2023-01-07
Fix multiple crashes when using VP9 depacketizer

aler9:
#### Description

Multiple optional fields of the VP9 RTP header are read regardless of the buffer size. This can easily lead to multiple crashes. This PR add the necessary checks in order to avoid crashes.

### DIFF
--- a/codecs/vp9_packet.go
+++ b/codecs/vp9_packet.go
@@ -207,12 +207,15 @@ func (p *VP9Packet) Unmarshal(packet []byte) ([]byte, error) {
 
 // Picture ID:
 //
-//      +-+-+-+-+-+-+-+-+
-// I:   |M| PICTURE ID  |   M:0 => picture id is 7 bits.
-//      +-+-+-+-+-+-+-+-+   M:1 => picture id is 15 bits.
-// M:   | EXTENDED PID  |
-//      +-+-+-+-+-+-+-+-+
+//	+-+-+-+-+-+-+-+-+
 //
+// I:   |M| PICTURE ID  |   M:0 => picture id is 7 bits.
+//
+//	+-+-+-+-+-+-+-+-+   M:1 => picture id is 15 bits.
+//
+// M:   | EXTENDED PID  |
+//
+//	+-+-+-+-+-+-+-+-+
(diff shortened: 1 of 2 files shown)


codecs/vp9_packet.go
```diff
@@ -313,24 +319,30 @@ func (p *VP9Packet) parseRefIndices(packet []byte, pos int) (int, error) {
...
+//	+-+-+-+-+-+-+-+-+                           -|
+//
 // N_G: |  T  |U| R |-|-| (OPTIONAL)                 .
-//      +-+-+-+-+-+-+-+-+              -|            . N_G times
-//      |    P_DIFF     | (OPTIONAL)    . R times    .
-//      +-+-+-+-+-+-+-+-+              -|           -|
 //
+//	+-+-+-+-+-+-+-+-+              -|            . N_G times
+//	|    P_DIFF     | (OPTIONAL)    . R times    .
+//	+-+-+-+-+-+-+-+-+              -|           -|
```
tmatth, line 345: Are all of these diagram formatting changes intentional? Some (like the added `//` lines) are inconsistent with the diagrames in the rest of the file.

tmatth:
@aler9 it looks like you [already pushed this to the v1 branch](https://github.com/pion/rtp/commit/c3ea6851e25b325c13e082006ea10b40e11f9a75), was that intentional?

codecs/vp9_packet.go
aler9: @tmatth that's just the result of `gofmt -w -s`, comments should be inserted in another way in order to avoid conflicts with standard tools (for instance, by using long comments with /* */ )

aler9:
@tmatth i didn't notice it but at the same time the v1 branch must be protected by an administrator in order to avoid conflicts. Anyway this PR is urgent as it allows anyone to crash any pion/webrtc-based software.

codecs/vp9_packet.go
tmatth: @aler9 ok since it landed in v1 already it'd make more sense to redo the diagrams with `/* */` in a subsequent PR then.

tmatth (approved):

tmatth:
@aler9 can you rebase?

PR #119, opened 2021-02-08
Refactor Packet struct to match RFC format

ffmiyo:
### Refactor Packet struct to match RFC format
Remove PayloadOffset from Header.
Don't keep unnecessary full data bytes inside Packet.
Packet Marshal and Unmarshal method API stay the same.

Fixes #90 

This PR do breaking changes as part of v2 plan. 
Related to #88 #79

PR #228, opened 2023-07-16
Fix API breakage on Header.Unmarshal

Sean-Der:
Return (int, error) and not just error

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -32,6 +32,7 @@ Rob Lofthouse <ri.lofthouse@gmail.com>
 Robin Raymond <robin@goheadroom.com>
 Sean DuBois <duboisea@twitch.tv>
 Sean DuBois <seaduboi@amazon.com>
+Sean DuBois <sean@pion.ly>
 Sean DuBois <sean@siobud.com>
 Simone Gotti <simone.gotti@gmail.com>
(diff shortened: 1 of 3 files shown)


Sean-Der:
@stv0g @aler9 This was the plan I believe? After this lands I can start tagging `v1` off of master branch again.

aler9:
@Sean-Der the change that you want to restore was part of the backward-incompatible PR #119, that was included in v1.7.0 and that i intentionally left out in my previous PR. Prior v1.7.0, signature of `Header.Unmarshal` was

https://github.com/pion/rtp/blob/4997f5536911192ea66dd8f47eae54934e9a1537/packet.go#L81

after v1.7.0, signature is

https://github.com/pion/rtp/blob/ccb8b11cd0efb85bef256e50a5b6b4c4ac74058a/packet.go#L78

Therefore, this PR restores compatibility with `v1.7.x` but breaks compatibility with `v1.6.5`. Probably it's better to merge it since version `v1.7.x` has surely more users than `v1.6.x`.

aler9 (approved):

Sean-Der:
@aler9 thanks for the explanation. What a mess :(

I will get some tooling setup to make sure this doesn’t happen again.

PR #227, opened 2023-06-12
Rollback to v1

aler9:
#### Description

As discussed in #221, this reverts the library to /v1 by including only backward-compatible changes, in order to restore the release cycle of this library, that has been stuck for over a year.

In particular, i did the following:

1. the library was reverted to the last commit before bumping to v2 (4e87540)
2. all non-breaking commits from the v1 branch were cherry-picked
3. all non-breaking commits from the master branch were cherry-picked
4. the following breaking commits from master were edited and merged:
    * #150 - this removed some structs, that have been re-added and marked as deprecated
5. the following breaking commits were left out and will be readded with a PR in a non-breaking form:
   * #119
   * #160

PR #234, opened 2023-07-28
Fix race condition in Packet.MarshalTo()

aler9:
#### Description

This was already fixed by #168 but got lost in #227. in SFUs, in order to distribute a packet to all clients, MarshalTo() is called in parallel by multiple routines, causing a race condition because the padding flag is dynamically set inside MarshalTo(). This is particular annoying when running automated tests. This PR fixes the issue by removing this write operation as discussed in #168.

### DIFF
--- a/packet.go
+++ b/packet.go
@@ -484,7 +484,6 @@ func (p Packet) Marshal() (buf []byte, err error) {
 
 // MarshalTo serializes the packet and writes to the buffer.
 func (p *Packet) MarshalTo(buf []byte) (n int, err error) {
-	p.Header.Padding = p.PaddingSize != 0
 	n, err = p.Header.MarshalTo(buf)
 	if err != nil {
 		return 0, err
--- a/packet_test.go
+++ b/packet_test.go
@@ -254,6 +254,7 @@ func TestBasic(t *testing.T) {
 				}},
 			},
 			Version:        2,
+			Padding:        true,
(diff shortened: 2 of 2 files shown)


Sean-Der (approved):

0x5e:
Hi @aler9 , Upgrade from 1.8.0 to 1.8.1, my h264 stream is broken. I have set `rtpHeader.Padding = true` before, and remove this made the stream ok again.
Is it required that we must not set the `rtpHeader.Padding` value?

aler9:
Eventually we could add a consistency check in MarshalTo that returns an error when Header.Padding is true and PaddingSize is zero.

0x5e:
Here's my part of code about send rtp packet:
```
	pkt := rtp.Packet{Header: rtp.Header{
		Version: 2,
		Padding:        true,
		Extension:      false,
		Marker:         (subindex == subcount),
		PayloadType:    videoMedia.Formats[0].PayloadType(),
		SequenceNumber: seq,
		Timestamp:      rtpTimestamp + index*RTP_VIDEO_TICK,
		SSRC:           ssrc,
		CSRC:           []uint32{0},
	}, Payload: buf.Bytes()}
	rtspClient.WritePacketRTP(videoMedia, &pkt)
```

I didn't set PaddingSize. the code works in 1.8.0, but not in 1.8.1.

PR #235, opened 2023-08-02
Export header extensions

renandincer:
#### Description

Exports RTP Header Extensions.

At Cloudflare we serialize RTP state for graceful upgrading of software using [tableflip](https://github.com/cloudflare/tableflip) as described at: https://blog.cloudflare.com/graceful-upgrades-in-go/. Exporting the extensions could be useful for others as well for similar reasons.

### DIFF
--- a/packet.go
+++ b/packet.go
@@ -11,8 +11,8 @@ import (
 
 // Extension RTP Header extension
 type Extension struct {
-	id      uint8
-	payload []byte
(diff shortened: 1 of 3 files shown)


aler9:
Hello, this is exactly the same as #233, personally i posted my opinion there.

jerry-tao:
Closed due to duplicated with #233 .

PR #239, opened 2023-09-20
Optimize the performance of H264 packaging

ziminghua:
Optimize the performance of H264 packaging

#### Description
Sending package H264 in my program is quite performance intensive. Then through ppfrof, it was found that the range of nalu is relatively slow, and it will create slices multiple times. Use bytes.Index lookup index can significantly improve performance

Before:
image

After:
image

### DIFF
--- a/codecs/h264_packet.go
+++ b/codecs/h264_packet.go
@@ -4,6 +4,7 @@
 package codecs
 
 import (
+	"bytes"
 	"encoding/binary"
 	"fmt"
 )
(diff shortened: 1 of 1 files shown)


Sean-Der:
Great change @ziminghua!

It is possible to support both start codes? https://learn.microsoft.com/en-us/windows/win32/directshow/h-264-video-types#h264-bitstream-with-start-codes

ziminghua:
Sorry, I didn't consider Appendix B originally, so I optimized the original performance. If I want to support both start codes, I think this is possible.

ziminghua:
@Sean-Der Hello, what do you think of the current update.

Sean-Der:
@ziminghua great work! Lets merge it. I am out of time, I will merge first thing tomorrow

Code is much easier to read also :)

Sean-Der (approved):

Sean-Der:
@ziminghua I tagged `v1.8.2` so you can use this now! Thank you so much for the contribution :)

PR #250, opened 2024-03-14
Add constants for know RTP payload types

dennwc:
#### Description
This PR adds constants for known RTP payload types based on [IANA registry](https://www.iana.org/assignments/rtp-parameters/rtp-parameters.xhtml). Docs for each type also references a relevant RFC.

`PayloadTypeFirstDynamic` constant tracks the first dynamic payload type. It can be used as an array size for static entries when handling RTP packets. 

Apart from static types that do not change, it adds a default value for a dynamic payload type (`PayloadTypeDefaultDynamic`). There's nothing special about it, it's just the one frequently seen in the wild.

#### Reference issue
none

### DIFF
--- a/payloads.go
+++ b/payloads.go
@@ -1,0 +1,70 @@
+// SPDX-FileCopyrightText: 2024 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+package rtp
+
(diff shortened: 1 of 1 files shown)


Sean-Der:
Cherry-picked with https://github.com/pion/rtp/commit/057eda35a6580f672dc827cfafae06adc5e151ac

PR #263, opened 2024-04-14
Fix crash in Vp8Packet.Unmarshal

jech:
We used to crash if the I bit was set and the payload was
one byte short.


jech:
Fixed in 14c61dc0359c379f52978826c66bfcd9d90f7ff7.  Sorry for the noise.

PR #264, opened 2024-04-15
AV1 samplebuilder support (WIP)

jech:
- Make AV1Packet implement Depacketizer
- Implement AppendToSample for AV1Packet

An attempt to implement the required hooks for the samplebuilder
to work with AV1.  This approach is likely to be both faster and
simpler than the one in #238.

COMPLETELY UNTESTED, DO NOT COMMIT.

### DIFF
--- a/codecs/av1_packet.go
+++ b/codecs/av1_packet.go
@@ -134,6 +134,8 @@ type AV1Packet struct {
 	// Each AV1 RTP Packet is a collection of OBU Elements. Each OBU Element may be a full OBU, or just a fragment of one.
 	// AV1Frame provides the tools to construct a collection of OBUs from a collection of OBU Elements
 	OBUElements [][]byte
+
+	videoDepacketizer
 }
 
 // Unmarshal parses the passed byte slice and stores the result in the AV1Packet this method is called upon
@@ -185,4 +187,10 @@ func (p *AV1Packet) Unmarshal(payload []byte) ([]byte, error) {
 	}
 
 	return payload[1:], nil
+}
+
+func (p *AV1Packet) IsPartitionHead(payload []byte) bool {
+	// at this point, p.Unmarshal may not have been called yet.  Check
+	// manually for Z == 0 & N == 1.
+	return len(payload) > 0 && (payload[0] & 0x88) != 0x08
 }
--- a/codecs/av1_packet.go
+++ b/codecs/av1_packet.go
@@ -194,3 +194,42 @@ func (p *AV1Packet) IsPartitionHead(payload []byte) bool {
 	// manually for Z == 0 & N == 1.
 	return len(payload) > 0 && (payload[0] & 0x88) != 0x08
 }
+
+// AppendToSample appends the contents of the packet to a sample.  It
+// reconstructs the size of the last OBU if required.
+func (p *AV1Packet) AppendToSample(sample, payload []byte) []byte {
+	if len(payload) < 1 {
+		return sample
+	}
+
+	switch p.W {
+	case 0:
+		sample = append(sample, payload...)
+	case 1:
+		sample = append(sample, obu.WriteToLeb128(uint(len(payload)))...)
+		sample = append(sample, payload...)
+	case 2:
+		l1, ll1, err := obu.ReadLeb128(payload)
+		if err != nil {
+			return sample
+		}
+		sample = append(sample, payload[:l1+ll1]...)
+		l2 := uint(len(payload)) - (l1 + ll1)
+		sample = append(sample, obu.WriteToLeb128(l2)...)
+		sample = append(sample, payload[l1+ll1:]...)
+	case 3:
+		l1, ll1, err := obu.ReadLeb128(payload)
+		if err != nil {
+			return sample
+		}
+		l2, ll2, err := obu.ReadLeb128(payload[l1+ll1:])
+		if err != nil {
+			return sample
+		}
+		l3 := uint(len(payload)) - (l1 + ll1 + l2 + ll2)
+		sample = append(sample, payload[:l1+ll1+l2+ll2]...)
+		sample = append(sample, obu.WriteToLeb128(l3)...)
+		sample = append(sample, payload[l1+ll1+l2+ll2:]...)
+	}
+	return sample
+}


alexpokotilo:
1)AppendToSample not called at all

2) func (p *AV1Packet) IsPartitionHead(payload []byte) bool  is called before func (p *AV1Packet) Unmarshal(payload []byte) ([]byte, error) { and hence you cannot use !p.Z && p.N and should use something like

	if len(payload) == 0 {
		return false
	}
	return (payload[0] & byte(0b10000000)) == 0

I was not able to test further as IsPartitionHead always false

jech:
As I've mentioned, this is completely untested.  However, do you see any good reasons why this approach shouldn't work?

jech:
Fixed.  However, as I've mentioned, this is completely untested, and I'd be very surprised if it worked.  The question is whether you agree that this could work in principle, since it is both more efficient and simpler than what you propose.

alexpokotilo:
> Fixed. However, as I've mentioned, this is completely untested, and I'd be very surprised if it worked. The question is whether you agree that this could work in principle, since it is both more efficient and simpler than what you propose.

AppendToSample implementation is wrong. There are maybe several OBUS in AV1Packet and AppendToSample should run the similar loop as func (p *AV1Packet) Unmarshal

But there is much simplier approach much better than my and yours: just return 

OBULen+OBU list right from Unmarshal
I checked and nobody cares about func (p *AV1Packet) Unmarshal. Who use AV1Packet without samplebuilder, they use AV1Packet for it's OBUelements map. They don't care about payload format at all.

All my fix was to workaround current useless av1 format. your AppendToSample introduce to keep current format but to produce right format after Unmarshal. Shouldn't we just return all we need from Unmarshal. Yes, we need to allocate for sample but your implementation do the same, but for only one OBU. This is wrong. There maybe multiple OBUs in Av1Packet.
I'd bet to return correct av1 bitstreama directly from Unmarshal

jech:
> AppendToSample implementation is wrong.

Interesting.

> There are maybe several OBUS in AV1Packet

That's right, and my implementation builds a sample as a sequence of LE128-prefixed OBUs.  Are you suggesting a different representation?

> I checked and nobody cares about func (p *AV1Packet) Unmarshal.

https://github.com/jech/galene/blob/master/codecs/codecs.go#L48

This code does the equivalent of AV1Packer.Unmarshal in an efficient manner.  I had to write it because Unmarshal was written by people who don't care about efficiency.

> Shouldn't we just return all we need from Unmarshal. 

Unmarshal is a low-leve function that should run in constant time and perform no allocations.  Check the H264 version of Unmarshal, it's done right, it doesn't try to parse the list of NALUs.  Please also read the comments in https://github.com/pion/rtp/pull/265 and https://github.com/pion/rtp/pull/266.

jech:
@alexpokotilo Another point is that you cannot reassemble fragmented OBUs in `Unmarshal`, since the packets haven't been reordered at this point: it's the samplebuilder's (or the jitter buffer's) job to reorder packets.  So there's really no alternative to doing the parsing later than at Unmarshal time.

alexpokotilo:
> @alexpokotilo Another point is that you cannot reassemble fragmented OBUs in `Unmarshal`, since the packets haven't been reordered at this point: it's the samplebuilder's (or the jitter buffer's) job to reorder packets. So there's really no alternative to doing the parsing later than at Unmarshal time.

you are right. I forget about fragmented OBUs

alexpokotilo:
> > AppendToSample implementation is wrong.
> 
> Interesting.
> 
> > There are maybe several OBUS in AV1Packet
> 
> That's right, and my implementation builds a sample as a sequence of LE128-prefixed OBUs. Are you suggesting a different representation?
> 

@jech, maybe my problem is that I see AppendToSample implementation but don't see its usage in your branch. Could you please finish your version so I can just use it and check it really support all cases.
Now I see a part of job. I don't understand how AppendToSample do the same OBUs processing as Unmarshal because you didn't add AppendToSample call to sample builder. If you call it in loop till all data is processed then ok, I got the point. If you call it without loop like you pointed in your https://github.com/pion/rtp/pull/238#issuecomment-2056512863 example I don't get the point then. Again speaking about remained fragment of last OBU. I don't see how you are going to process it.

Long story short. You are right and my [version](https://github.com/pion/rtp/pull/238)  is not ideal. I don't like it either.
When I did it I tried to close the gap in AV1Packet implementation.  I had to use AV1Frame to correctly process last remained OBU as AV1Packet drops OBUElements on unmarshal and don't safe last not complete OBU. For example H264Packet save fragmented nal for further processing. AV1Packet doesn't do that so your AppendToSample will not work.

From performance pov it doesn't matter where you process OBUs: in unmarshal method or in AppendToSample right next to it.
I can see AV1Packet is used only in samplebuffer and some tests. 
if we use AppendToSample or return processed data from unmarshal the total performance will be the same. You will not be able to eliminate allocation in unmarshal if you add fragmented obu support there. Take a look H264Packet. If you have fuaBuffer from previous packet then "p.fuaBuffer = append(p.fuaBuffer, payload[fuaHeaderSize:]...)" performaed. and this is allocation.

Now AV1Packet just drop OBUElements on every unmarshal. it doesn't care about previous fragments. I had to use AV1Frame to combine OBUs from AV1Packets as AV1Packet doesn't care about it.

What you propose is :
- add new interface  AppendToSample
- add check of it into sample buffer
- call this AppendToSample in loop.
And all of this because you don't want allocations in AV1Packets's unmarshal. Maybe you are right and unmarshal should do allocations but look at H264Packet unmarshal. It does allocation. and it handles tail from previous unsharshal itself. So it's not just av1 problem after all. To be honest I don't see any problem with this approash at all. If you want to check just rtp headers you can just process packet payload or we can add new method to all Packets like unmarshalHeader. we can use it in unmarshal as well.

Don't get me wrong please. I admit my initial approach is wrong and I'm starving to test your version but it doesn't exist.

I still don't see a problem in adding tail support into AV1Packets the same way H264Packet do and return necessary format from it without adding new interface AppendToSample. at the end of the day you will have to allocate somewhere. Why not in unmarshal ? What so special in this method so that you don't want to do the job there? these Packet structures are used mainly in samplebuilder. You don't add any gain moving processing in new method AppendToSample.

If you complain about unmarshal-only approach only from design POV - I'm sorry I don't completely confident to speak from design POV. I'm not Packet approach designed. But If you thing your approach gain something to samplebuffer users - please complete it and I will create performance test to check where you are right or not.

I think it's time to  complete AppendToSample so I can do tests and check whether it works or not. Now I cannot and don't see how I can help to prove your points or argue with them.

My main problem with AppendToSample approach is that I don't get what is so special with av1 so we need to handle it different way from h264? do we have tails in h264? yes we do. How we process them? in unmarshal. Do we do allocation in h264's unmarshals? Yes, in case we have previous tail. Doesn't h264' unmarshal return ready to append buffer. Yes it is.

You want to have lightweight header parser without all body processing - go ahead but this is a different story. 
I don't see how you improve performance by moving processing into AppendToSample and call it right after  unmarshal. What samplebuffer users gain from this ?

Do I agree that AV1Packet made in different way than others Packets? Yes.
Should we add tail support in it to not rely on Av1Frame for assembling? 100%
Should we add AppendToSample to offload unmarshal? For what ?
Should we remove OBUElements processing? Well yes, but Av1Frame users will not be happy right now. this is legasy, We can add flag like @Sean-Der proposed in one of your pulls.

This is all I can say about it.

alexpokotilo:
> @alexpokotilo Another point is that you cannot reassemble fragmented OBUs in `Unmarshal`, since the packets haven't been reordered at this point: it's the samplebuilder's (or the jitter buffer's) job to reorder packets. So there's really no alternative to doing the parsing later than at Unmarshal time.

Please look at H264Packet. it does process tails from previous unmarshal call relying on the fact that samplebuffer call unmarshal in correct order. This is what I see in the code at least. Check func (p *H264Packet) Unmarshal(payload []byte) ([]byte, error) and it's p.fuaBuffer processing. If unmarshal doesn't rely on packet order this approach would not work for h264.

jech:
> Please look at H264Packet. it does process tails from previous unmarshal call relying on the fact that samplebuffer call unmarshal in correct order. 

Yes, and that's wrong, for at least three reasons:

  * people expect Unmarshal and Marshal should be (at least logically) inverses; if you defragment at Unmarshal time, you loose this property, which causes subtle bugs;
  * there is no place where it is documented that Unmarshal must be colled in sequential orde (the VP8 and VP9 formats don't require it), so the requirement causes subtle bugs;
  * there are applications that want to call Unmarshal just to check header flags, these applications do not want to pay the whole price of parsing the OBU or NALU list.

Alex, let us please do the right thing in AV1, and let's fix H.264.  I've started here:

  * https://github.com/pion/rtp/pull/266
  * https://github.com/pion/rtp/pull/268
  * https://github.com/pion/rtp/pull/267

alexpokotilo:
> > Please look at H264Packet. it does process tails from previous unmarshal call relying on the fact that samplebuffer call unmarshal in correct order.
> 
> Yes, and that's wrong, for at least three reasons:
> 
>     * people expect Unmarshal and Marshal should be (at least logically) inverses; if you defragment at Unmarshal time, you loose this property, which causes subtle bugs;
> 
>     * there is no place where it is documented that Unmarshal must be colled in sequential orde (the VP8 and VP9 formats don't require it), so the requirement causes subtle bugs;
> 
>     * there are applications that want to call Unmarshal just to check header flags, these applications do not want to pay the whole price of parsing the OBU or NALU list.

So maybe add Another method to help these application and keep current interface to others ?
I respect your passion and willing to change things you consider not right. I just don't see anything wrong with the way it is for h264 and others. I vote to fix AV1Packet. If you can convince maintainers then you are right. I don't see any reason for this. But I could be wrong

aleksitto-gh:
Hi guys, thanks for this topic, it would be nice to have AV1 support. 

When I try to send encoded AV1 video via webrtc - I see the following in my browser logs (probably related to this task):

```
[63288:64259:0701/173544.298704:WARNING:video_rtp_depacketizer_av1.cc(313)] Mismatch in obu_size. signaled: 0, actual: 2350
[63288:64259:0701/173544.332409:WARNING:video_rtp_depacketizer_av1.cc(313)] Mismatch in obu_size. signaled: 0, actual: 3088
```
(video is not playing... )

According to the browser source code something wrong with the OBU size calculation:

```
obu_info->payload_size = 
    obu_info->data.size() - obu_info->prefix_size - size_of_obu_size_bytes;

  if (obu_size_bytes != obu_info->payload_size) {
    // obu_size was present in the bitstream and mismatches calculated size.
    RTC_DLOG(WARNING) << "Mismatch in obu_size. signaled: " << obu_size_bytes
      << ", actual: " << obu_info->payload_size;
    return false;
  }
```

If you have any ideas on how to fix it - it would be very nice

jech:
Yes, I'm planning to work on it.  I'm actually planning a fairly comprehensive rework of the packet code:

- add a new interface, so that the zero-alloc mode has feature-parity with the (currently default) wasteful mode;
- modify the sample-builder to work with the zero-alloc mode;
- add support for the new interface to the AV1 codec (zero-alloc mode only, since I'm planning to deprecate the wasteful mode);
- convince Sean to make zero-alloc the default (incompatible change).

Unfortunately, I won't be able to work on that before the end of October.

lebedyncrs:
> Yes, I'm planning to work on it. I'm actually planning a fairly comprehensive rework of the packet code:
> 
> * add a new interface, so that the zero-alloc mode has feature-parity with the (currently default) wasteful mode;
> * modify the sample-builder to work with the zero-alloc mode;
> * add support for the new interface to the AV1 codec (zero-alloc mode only, since I'm planning to deprecate the wasteful mode);
> * convince Sean to make zero-alloc the default (incompatible change).
> 
> Unfortunately, I won't be able to work on that before the end of October.

Sounds exciting, please let me know if you would need help with testing. Does this PR is going to solve https://github.com/pion/rtp/issues/273 ?

joeturki:
Hello, I'm going to start working on AV1 support next, can I adopt this?  \:)

jech:
No, but we can work together.

joeturki:
I meant asking if I can continue your work; I have time to work on it, and I do agree with your approach here \:) can you share with me with me what you found with this experiential and the rest of your vision? fixing other codecs seems interesting but we must do it in a breaking way.

jech:
Currently, the samplebuilder does that:
```
    buf, err := s.depacketizer.Unmarshal(packet.Payload)
    data = append(data, buf)
```
At first sight, that looks good, but in practice the Depacketizer interface is implemented by the various packet types, which is wrong, since it confuses, say, XXXPacket.Unmarshal and Depacketizer.Unmarshal.  But from the point of the library user the two functions have different usage patterns:

  * XXXPacket.Unmarshal is a stateless function (we shouldn't have any state in XXXPacket), that may be called in any order (you don't expect to be required to call XXXPacket.Unmarshal in sequential order, for example in receive order);
  * Depacketizer.Unmarshal is a stateful function, that encodes the dependencies between packets in the Depacketizer.

An example of the confusion is here: https://github.com/pion/rtp/blob/master/codecs/h264_packet.go#L186.  The H264Packet contains state (fuaBuffer) which the Unmarshal method uses to carry dependencies between packets.  The effect is that if the H264packet.Unmarshal method is called non-sequentially (say, in reception order rather than in seqno order), the data will become corrupted.  We'll hit the same issue with AV1.  (VP8 and VP9 have simple packetisation rules that don't require state.)

The solution is to add a method `GetDepacketizer` to XXXPacket, and pass a Depacketizer rather than an XXXPacket to the SampleBuilder function.  The XXXPacket.Unmarshal functions need to be made stateless and efficient, which is what I've started doing when the NoAlloc flag is set.

Finally, as long as we're breaking compatibility, we should replace the Depacketizer.Unmarshal function with

    func (d *Depacketizer) AppendData(data) error

which does the equivalent of

    data = append(data, d.Unmarshal())

This way, the SampleBuilder can call AppendData instead of Unmarshal, avoiding one allocation.

joeturki:
> XXXPacket.Unmarshal is a stateless function (we shouldn't have any state in XXXPacket), that may be called in any order (you don't expect to be required to call XXXPacket.Unmarshal in sequential order, for example in receive order);

As a user of the library this confused me with the h264 library.

> The solution is to add a method GetDepacketizer to XXXPacket, and pass a Depacketizer rather than an XXXPacket to the SampleBuilder function. The XXXPacket.Unmarshal functions need to be made stateless and efficient, which is what I've started doing when the NoAlloc flag is set. 
> Finally, as long as we're breaking compatibility, we should replace the Depacketizer.Unmarshal function with

I don't think we can break it now, maybe we add a new interface and make it work with the new interface and the old Unmarshal  / Marshal. This way we can add to `av1` and even other codecs (without removing the Unmarshal  / Marshal, just soft deprecate) Until we change that in the next major version in the future.

What do you think?

jech:
I think we should first implement the right thing on a separate branch, without regard to backwards compatibility.  Once we're agreed that's the right thing, we see how much nasty hacks we can accept in order to preserve compatibility.

joeturki:
> I think we should first implement the right thing on a separate branch, without regard to backwards compatibility. Once we're agreed that's the right thing, we see how much nasty hacks we can accept in order to preserve compatibility.

I'm down for this and I have the time and the energy to do both, but first I think I will have to ship it to this repo, I think your approach is good, and we can do type check. But do you mean i should ship it with marshal / unmarshal, And we make the new `Depacketizer` interface for the new repo?

joeturki:
maybe we can just update the Depacketizer interface to this and leave the `Unmarshal` function for now but with soft deprecate and  remove it by the next major release?

```go
type Depacketizer interface {  
    // soft deprecated, just for end users.
    Unmarshal(packet []byte) ([]byte, error)  

    // what pion will use.
    AppendData(dest *[]byte, packet []byte) error  

    IsPartitionHead(payload []byte) bool  

    IsPartitionTail(marker bool, payload []byte) bool  
}  
```

what do you think @jech ?

joeturki:
> I think that for now we can work with the current rtp.Depacketizer interface, we just need to decouple XXXPacket from XXXDepacketizer for all codecs with non-trivial depacketisation rules.

Okay i see, so do i make a PR for Unmarshal for `av1`? I will add an option to disable parsing like you did with h264.

joeturki:
> Why continue with the broken approach? Wouldn't it be better to start with doing the decoupling for H264Packet?

The only reason is that I don't want to break anything for the end user \:) and I want the current users to be able to use `av1` with this repo. But i think we should do this "rewrite" by the next major Pion release.

Maybe the two interfaces approach is good? I want to ship av1 and fix all the reported bugs related to av1 by this week.

If you found a good way to do all of this without breaking any changes, just provide your design approach and i will ship the implementation in no time \:)

joeturki:
> Once that is done, we can start adding non-broken depacketizers to all of the XXXPacket types.

Okay I think this is good, the only issue if users provide their own `Depacketizer` is this a thing? I can do a quick code search.

joeturki:
Okay, So I'm not really sure, I think your approach is obviously better for `av1`, but I'm not sure if we can break changes, I think we should post it to the slack and see what people see. then i will see what route we go, Again i believe i can do the implementation itself in no time, but I just need other contributors to agree :)

What do you think?

jech:
In my experience, getting others to agree is easier when you have working code to show them.

joeturki:
Well, my point is that I don't want to waste time that can be spent on other issues :) 

but let's say we can break changes what do you think the final design should look like?

jech:
```go
sb := depacketizer.New(... AV1Packet{}.GetDepacketizer(), ...)
var p AV1Packet
for {
    n, err := track.Read(buf)
    p.Unmarshal(buf, n)
    depacketizer.Push(&p)
    for {
        ...
        depacketizer.PopWithTimestamp()
        ..
}

The only change wrt. what is done now is the call to `depacketizer.New`, which requires the caller to call `GetDepacketizer` by themselves.  For now, we keep that optional for existing codecs, but require it for AV1, so that people will naturally migrate to the new API when they implement AV1 support.

**Comment by jech** [comment]
> I think we should post it to the slack 

I think it's a good idea to send a pointer to this discussion to Slack.  I'll try to check if there's any discussion on Slack, but no promises, I'm developing Slack-phobia.  (What's wrong with mailing lists?)

**Comment by joeturki** [comment]
Okay i posted it to the slack, we will see by tomorrow, I'm going to work on other issues today \:)

**Comment by joeturki** [comment]
@jech I was thinking of shipping a version of the AV1 Depacketizer with the current unmarshal API. While I agree it's not ideal, I can expose all the necessary APIs to ensure your refactor works seamlessly, or for you to us it this way.

After that, I'll create a new Depacketizer for sampleBuilder using your proposed approach. I'll also upgrade AV1 to work with it and update all the codecs to support the new API. I believe I can complete all of this by the end of January. This way, the new API will be opt-in for now, so we don't break anything.

And we can add a depeacture flag, and remove `unmarshal` depacketizers by the next major release.

Let me know what you think?

**Comment by jech** [comment]
Joe, I think it's a very bad idea.  AV1 packets cannot be parsed in isolation, they need to be parsed in increasing seqno order.  If you use the current API for AV1, you require that packets are parsed in seqno order, and your code will quite simply be incorrect in the case of packet reordering.

I understand the need to keep broken code around for compatibility reasons.  I don't understand why you want to be writing new code with the full knowledge that it's incorrect.

**Comment by Sean-Der** [comment]
@jech What do you think about this falling under a 'separation of concerns'?

We wan't to allow a user to run a broken AV1 stream through pion/rtp. If you have a packet capture on disk and you just want best effort.

If you are dealing with a network you should put unmarshal after a interceptor.JitterBuffer and we only want to have that code in one place.

**Comment by Sean-Der** [comment]
This issue has been open for over a year. I think it would be exciting to have it closed out. Even if the API isn't perfect, better than having nothing!

Unless you think we are shipping something that is worse than having nothing. If it creates a situation that is harmful/insecure.

**Comment by joeturki** [comment]
> The API that I've outlined is actually simpler than the current API. I'd like to understand why Joe prefers to implement an API that is both more confusing and not correct.

No, that's not what I mean :) All I'm saying is that I want to implement AV1 support for the current sampleBuilder API so we have `av1` support, even if it's not perfect, but It's something. Then, we can work on designing a new sampleBuilder API separately, and we can call it something else, I think both can coexist. and I can work on both \:)

**Comment by joeturki** [comment]
> And yes, it makes things worse: if AV1Packet.Unmarshal does non-trivial processing, then any application that uses AV1Packet.Unmarshal just to get at the header flags will become slower. Which is just one of the reasons why we should decouple XXXPacket from XXXDepacketizer.

I agree, and I can create additional APIs in the av1 library for use cases like this \:) until we fix samplebuilder systematically

**Comment by jech** [comment]
> ll I'm saying is that I want to implement AV1 support for the current sampleBuilder API 

Once again: the SampleBuilder API does not change: the SampleBuilder alread takes a Depacketizer as argument:

  https://github.com/pion/webrtc/blob/master/pkg/media/samplebuilder/samplebuilder.go#L62

The issue is with having the same structure implementing Packet and Depacketizer: if AV1Packet implements rtp.Depacketizer, then there is state in AV1Packet, and that has a number of negative consequences.

In order to avoid these consequences, it is enough to have two structures: AV1Packet, which is stateless and implements Unmarshal efficiently, and AV1Depacketizer, which contains state and does any stateful packet processing that's required.

Once again: the SampleBuilder API is correct, it already cleanly distinguishes Packet from Depacketizer.  The guilty party is H264Packet, which implements the Depacketizer interface rather than having a distinct H264Depacketizer.  Please do not repeat the error of H264Packet, do not put any state that depends on previous packets in AV1Packet, put it in a separate struct AV1Depacketizer.

**Comment by joeturki** [comment]
> The issue is with having the same structure implementing Packet and Depacketizer: if AV1Packet implements rtp.Depacketizer, then there is state in AV1Packet, and that has a number of negative consequences.

Okay, I can do this :) Could you suggest a clean design to implement it without breaking existing code or mistakenly invoking the sample builder on the AV1Packet? like where would we wire it to call "getDepacketizer" if it's an av1 packet?

**Comment by jech** [comment]
> breaking existing code 

We're not breaking existing code since the SampleBuilder does not currently work with AV1, so any code that attempts to do that is already broken.  That's why this is a unique opportunity that we must not miss.

> mistakenly invoking the sample builder on the AV1Packet

Don't implement the PartitionHead functions on AV1Packet, and it won't satisfy the Depacketizer interface.

> where would we wire it to call "getDepacketizer"

Just a method of AV1Packet that does not use its parameter (Go does not have static methods).

```
    func (_ *AV1Packet) NewDepacketizer() *AV1Depacketizer
```

joeturki:
Okay, I think we good, Can I start working on this? I will make a new branch, and i will try to ship it this week, I will make it religiously up to the spec. But I will fix the lint issues for this repo first.

We good? :)

jech:
> I will make it religiously up to the spec.

Don't.  Just keep in mind the requirement that AV1Packet.Unmarshal should be cheap and stateless.

joeturki:
I think the main issue here is separating `AV1Packet` and the av1 `Depacketizer`. 

> Don't feel overly bound by my strawman design

I'm not sure about using NewDepacketizer as a method on AV1Packet; where would the state be stored? I think a Depacketizer with a `consume(av1packet)` method might be a better approach. 

I'll create an issue and explore all possible designs after I fix the lint for this repo.

I'm happy this is resolved, thanks :)

joeturki:
I see.

JoeTurki:
closed in favor of https://github.com/pion/rtp/pull/294

PR #265, opened 2024-04-15
remove av1 obu parsing

jech:
- Remove OBU parsing from AV1Packet
- Remove codecs/av1/frame/av1.go

Remove OBU parsing from AV1Packet
    
In Pion, the Unmarshal operation is supposed to be
fast and perform no allocation, as it is sometimes used to check
just a single flag in the packet header.  The AV1Packet
implementation breaks this property by parsing the whole list
of OBUs at Unmarshal time, even though it is likely to be unneeded.
    
We remove OBU parsing from Unmarshal, to make the method consistent
with the other implementations in Pion (for example, H264Packet
does not parse the list of NALs at Unmarshal time).  We also remove the code in
codecs/frame, which depends on the OBU parsing.

If the code in codecs/frame is required, I volunteer to put it back in a way that
doesn't require parsing the whole list of OBUs at every call to Unmarshal.


Sean-Der:
Sorry about that @jech 

Could we make it opt-out?

jech:
> Could we make it opt-out?

Yes, that's what I'm suggesting.  I suggest adding a new function:

    func (p *AV1Packet) ParseOBUs() ([][]byte, error)

so that callers can opt-in simply by calling it.

Sean-Der:
What do you think of adding a `NoParse` member to `AV1Packet` ? Or something with different naming. It would be nice if we can avoid breaking the public API for existing users. 

All the Unmarshal functions have the same problem. They keep getting expanded to add more functionality.

jech:
> What do you think of adding a NoParse member to AV1Packet ? 

It means that the default path is the inefficient behaviour, which has a bad smell.  I'll do it, but only if you insist really strongly.

Sean-Der:
Replaced by https://github.com/pion/rtp/pull/266

PR #266, opened 2024-04-15
make av1 obu parsing optional

jech:
- Make parsing of the OBU list in AV1Packet optional
- Avoid usage of OBUElements in av1_packet_test.
- Mark ReadFrames as being deprecated.

In Pion, the (*XXXPacket).Unmarshal operation is supposed to be
fast and perform no allocation, as it is sometimes used to check
just a single flag in the packet header.  The AV1Packet
implementation breaks this property by parsing the whole list
of OBUs at Unmarshal time, even though it is likely to be unneeded.

Note that I strongly object to this solution, since it makes the
inefficient path (both in space and in time) the default, and thus
will make people write inefficient code.  However, if that's the
cost to getting the brain-damange reverted, I won't scream too
loudly.


jech:
I've just made the accessor into a method, so its presence can be tested by callers (by checking whether the packet implements a given interface).  I see that H264 and H265 have the same problem, so I've made the name codec-agnostic.

Sean-Der (approved):

PR #267, opened 2024-04-15
Test that Unmarshal doesn't do any allocations

jech:
The various Unmarshal functions are low-level functions that
should not do any extensive parsing.  Anything that requires
allocations should be done in a user-callable function so that
applications can call Unmarshal without concern for efficiency.


jech:
This currently fails for H264, H265 and AV1.  If either #265 or #266 gets applied, it will pass for AV1.  Once that happens, I'll fix H264 and H265.

jech:
If #266 and #268 are applied, this passes for AV1 and H.264.  I'm not going to fix H.265, since I'm not interested in single-platform codecs.

Sean-Der (approved):

PR #269, opened 2024-04-27
Fix VP9 decoding on iOS

aler9:
#### Description

The current implementation of the VP9 payloader produces payloads that are not compatible with iOS. This is because the payloader provides only the muxing strategy called "flexible mode".

According to the VP9 RFC draft, there are two ways to wrap VP9 frames into RTP packets: the "flexible mode" and the "non-flexible mode", with the latter being the preferred one for live-streaming applications. In particular, all browsers encodes VP9 RTP packets in the "non-flexible mode", while iOS supports decoding RTP packets in this mode only, and this is probably a problem shared by other implementations.

This patch improves the VP9 payloader by adding support for the "non-flexible mode". The "flexible mode" is retained and a flag is provided to perform the switch between the two modes.

#### Reference issue

https://github.com/bluenviron/mediamtx/issues/3060

### DIFF
--- a/codecs/vp9/bits.go
+++ b/codecs/vp9/bits.go
@@ -1,0 +1,65 @@
+// SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+package vp9
+
+import "errors"
+
+var errNotEnoughBits = errors.New("not enough bits")
+
+func hasSpace(buf []byte, pos int, n int) error {
+	if n > ((len(buf) * 8) - pos) {
+		return errNotEnoughBits
+	}
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


Sean-Der:
Thank you @aler9! I am in full support of this, happy to review w/e you are ready.

If Flexible mode doesn't have value (and is a burden) we could drop it.

aler9:
Hello @Sean-Der, now everything is ready for review, thanks for being available.
Regarding flexible mode, i don't know whether it has a value or not, keeping it is not a burden and avoids breakages in existing implementations.

Sean-Der:
LGTM @aler9! Thank you so much :)

Sean-Der (approved):

PR #270, opened 2024-05-01
Add an index check to prevent out of index

HyeockJinKim:
#### Description
In the current code, `binary.BigEndian.Uint16` is causing an out of index panic while traversing the for loop.
Add an length check to prevent panic from occurring.


at-wat:
@HyeockJinKim thank you for the fix!
Could you add a simple unit test that reproduces the panic on the original code?

HyeockJinKim:
I added 2 test cases.

HyeockJinKim:
Can I get a review again? @at-wat

codecs/h264_packet_test.go
```diff
@@ -189,6 +192,20 @@ func TestH264Packet_Unmarshal(t *testing.T) {
 	} else if !reflect.DeepEqual(res, singlePayloadMultiNALUUnmarshaledAVC) {
 		t.Fatal("Failed to unmarshal a single packet with multiple NALUs into avc stream")
 	}
+
+	res, err = pkt.Unmarshal(singlePayloadWithBrokenSecondNALU)
+	if err != nil {
+		t.Fatal(err)
```
at-wat, line 198: Is it expected not to return error when tried to unmarshal a broken packet?

HyeockJinKim: When receiving the nalus of STAP-A, I was thinking of returning only the parsed nalus rather than returning an error because of the multiple nalus received.
When receiving a corrupted packet, is it better to return an error rather than returning the parsed nalus?

at-wat: I'm not very familiar with h264.
Is the result data usable even if some NALUs are missing?
Also, is it fine without notifying the caller that the packet is corrupted?

HyeockJinKim: In H.264, STAP-A packets can contain multiple nalu within an rtp packet. (A nalu can be an I Frame, P Frame, SPS, PPS, etc.).

If the H264 Unmarshal fails, returning an error seems to prevent decoding the wrong packet. (It also lets the end user know that there was a wrong packet.)
If it does not return an error, but only the parsed NALU, then if there were bad packets, it would not decode them anyway, and if the missing NALU was not essential for decoding, it would decode H264 packets.

In `ffmpeg`, which is used as a third_party in `chromium`, the STAP-A decode process parses as many nalus as it can, and then returns only the parsed nalus. Instead, it returns an error when not even a single nalu is parsed.
https://source.chromium.org/chromium/chromium/src/+/main:third_party/ffmpeg/libavformat/rtpdec_h264.c;l=222-223?q=stap-a&ss=chromium

In my opinion, it would be nice if pion would decide how to handle this.

at-wat: @Sean-Der @gqgs what do you think?

HyeockJinKim: Just in case you've forgotten, here's a reminder. @Sean-Der @gqgs

at-wat: I asked @Sean-Der on Slack to check this out

Sean-Der (approved):

Sean-Der:
Sorry for the delay, this LGTM!

I agree on returning the partial. It would be nice if we could log, but not available unfortunately.

PR #274, opened 2024-07-06
Fix RTP padding length validation

sirzooro:
Added validation of RTP padding length in received packets. Also check for zero padding length when marshaling.

### DIFF
--- a/error.go
+++ b/error.go
@@ -21,4 +21,6 @@ var (
 	errRFC8285TwoByteHeaderSize    = errors.New("header extension payload must be 255bytes or less for RFC 5285 two byte extensions")
 
 	errRFC3550HeaderIDRange = errors.New("header extension id must be 0 for non-RFC 5285 extensions")
+
+	errInvalidRTPPadding = errors.New("invalid RTP padding")
 )
--- a/packet.go
+++ b/packet.go
@@ -215,6 +215,9 @@ func (p *Packet) Unmarshal(buf []byte) error {
 
 	end := len(buf)
 	if p.Header.Padding {
+		if end <= n {
+			return errTooSmall
+		}
 		p.PaddingSize = buf[end-1]
 		end -= int(p.PaddingSize)
 	}
@@ -475,6 +478,10 @@ func (p Packet) Marshal() (buf []byte, err error) {
 
 // MarshalTo serializes the packet and writes to the buffer.
 func (p *Packet) MarshalTo(buf []byte) (n int, err error) {
+	if p.Header.Padding && p.PaddingSize == 0 {
+		return 0, errInvalidRTPPadding
+	}
+
 	n, err = p.Header.MarshalTo(buf)
 	if err != nil {
 		return 0, err


boks1971 (approved):
LGTM

aler9:
Hello, although the specification says "If the padding bit is set, the packet contains one or more additional padding octets", I don't think it's a good idea to add a strict check on the topic since there are lots of devices that are not behaving accordingly. The only result would be decreasing the compatibility level of this library.

sirzooro:
@aler9 I have removed second check. Its first part checked for zero length padding, what is allowed now, Second part was redundant, the same check was done below after closing curly bracket. I left one check to make sure that there is at least one byte after header. Is it OK now?

aler9:
the remaining check, which is

https://github.com/pion/rtp/blob/cdd339011d96fa6a2f2200e5661d0a8b355cea11/packet.go#L218-L220

is almost a duplicate of a check that is present 4 lines below:

https://github.com/pion/rtp/blob/cdd339011d96fa6a2f2200e5661d0a8b355cea11/packet.go#L224-L227

The only difference is that it doesn't allow a zero-sized padding, again decreasing compatibility.

At the same time, the check on marshal:

https://github.com/pion/rtp/blob/cdd339011d96fa6a2f2200e5661d0a8b355cea11/packet.go#L481-L483

Prevents from marshaling invalid packets, which is something that not everyone wants to do, since a user might want to route incoming invalid packets to other sources by unmarshaling and marshaling them.

Therefore, this PR adds mandatory compliance checks which are not particularly useful and furthermore it decreases performance, since the additional lines in Unmarshal() are run during each packet unmarshaling, which is a mandatory operation that might happen hundreds of times per second for each video stream.

If you want to add compliance checks on RTP packets, you are free to do it directly inside your software rather than in the base library.

By the way, this is just may opinion which is in no way a final judgement.

sirzooro:
> the remaining check, which is
> 
> https://github.com/pion/rtp/blob/cdd339011d96fa6a2f2200e5661d0a8b355cea11/packet.go#L218-L220
> 
> is almost a duplicate of a check that is present 4 lines below:
> 
> https://github.com/pion/rtp/blob/cdd339011d96fa6a2f2200e5661d0a8b355cea11/packet.go#L224-L227
> 
> The only difference is that it doesn't allow a zero-sized padding, again decreasing compatibility.

This remaining check verifies that field with padding length is not part of RTP header. According to RTP RFC "The last octet of the padding contains a count of how many padding octets should be ignored, including itself". In other words. it is not possible to create zero-length padding by setting padding length to zero - you have to set padding bit in RTP header to 0.

> At the same time, the check on marshal:
> 
> https://github.com/pion/rtp/blob/cdd339011d96fa6a2f2200e5661d0a8b355cea11/packet.go#L481-L483
> 
> Prevents from marshaling invalid packets, which is something that not everyone wants to do, since a user might want to route incoming invalid packets to other peers by unmarshaling and marshaling them.

Attempt to do this causes crash in MarshalTo, I added this check to prevent it.

> Therefore, this PR adds mandatory compliance checks which are not particularly useful and furthermore it decreases performance, since the additional lines in Unmarshal() are run during each packet unmarshaling, which is a mandatory operation that might happen hundreds of times per second for each video stream.
> 
> If you want to add compliance checks on RTP packets, you are free to do it directly inside your software rather than in the base library.
> 
> By the way, this is just my opinion which is in no way a final judgement.

PR #277, opened 2024-08-06
Fix rare SRTP loss decode failure

kcaffrey:
As described in https://webrtc-review.googlesource.com/c/src/+/358360 there can be a problem when the sequence number starts near the rollover point and there is packet loss. As linked to in that issue, libsrtp recommends having the starting sequence number be less than 2^15 to avoid that problem.

### DIFF
--- a/sequencer.go
+++ b/sequencer.go
@@ -4,7 +4,6 @@
 package rtp
 
 import (
-	"math"
 	"sync"
 )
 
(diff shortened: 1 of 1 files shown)


Sean-Der (approved):

Sean-Der:
Thank you @kcaffrey merge + tag where ever you need it :)

boks1971:
Good catch!

nils-ohlmeier:
@kcaffrey you beat me to it :-)
I think this is worth releasing a new version, because without this fix you will get random call failures.

PR #284, opened 2024-11-29
fix issue with Packet.PaddingSize

strangesast:
memory from previous calls to Unmarshal is not correctly reset with zero value.  this causes issues when decoding packets into the same packet interface and causes payloads to be decoded incorrectly

Screenshot 2024-11-26 at 9 33 56 AM


Sean-Der:
Amazing, thank you so much @strangesast sorry this took so long :(

Sean-Der (approved):

PR #285, opened 2024-12-25
H265 payloader

z1rachl:
#### Description

Add h265 payloader support, that allows to split large h265 packets to FUs not bigger than requested mtu. It is mostly based on h264 payloader with rtp headers changes according to [RFC 7798](https://datatracker.ietf.org/doc/html/rfc7798)


Sean-Der:
Hi @z1rachl would you not like to merge this anymore?

PR #165, opened 2022-02-09
Add H265 payloader and depacketizer

kevmo314:
This change completes the H265 implementation.

PR #286, opened 2024-12-25
add h265 payloader

z1rachl:
#### Description

Add h265 payloader support, that allows to split large h265 packets to FUs not bigger than requested mtu. It is mostly based on h264 payloader with rtp headers generation changes according to [RFC 7798](https://datatracker.ietf.org/doc/html/rfc7798)

### DIFF
--- a/codecs/h265_packet.go
+++ b/codecs/h265_packet.go
@@ -31,7 +31,95 @@ const (
 	h265NaluFragmentationUnitType = 49
 	// https://datatracker.ietf.org/doc/html/rfc7798#section-4.4.4
 	h265NaluPACIPacketType = 50
+
+	h265FuaNALUType = 49
+	h265VpsNALUType = 32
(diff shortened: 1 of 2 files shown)


Sean-Der:
This looks perfect! I will fix linter and merge tonight :)

excited to use ir

kawaway:
@z1rachl
Thanks for the response. I have confirmed that I can play it in Safari using #287 with minor bug fixes from #165, Chrome is still checking. (I am checking which future flag to enable).

y-kawawa:
We have confirmed that it works on chrome in #287 based on #165. So functionality and whether #165 still has the motivation to move forward with PR will be the determining factor.
スクリーンショット 2025-01-09 12 56 01

joeturki:
Closed in favor of https://github.com/pion/rtp/pull/287

PR #289, opened 2025-01-10
Upgrade golangci-lint, more linters

joeturki:
#### Description

This is the first PR of two I fixed all the issues that doesn't require refactor, ignored the ones that does.

https://github.com/pion/.goassets/pull/220


joeturki (commented):
LGTM

PR #291, opened 2025-02-09
Implement av1 depacketizer

JoTurk:
#### Description

Adds AV1Depacketizer which implements the Depacketizer interface for AV1

NOTE: I believe this meets the spec, but I'm unsure how it will interact with the sample builder. I’ve only tested it in Chrome, where it seems to work, but I need to conduct more thorough testing before we merge. I'll set up a test environment for AV1 with different encoders and clients, run end-to-end tests on Pion, and also fix the AV1 Packetizer (I have a list of a few bugs to address).

#### Reference issue

https://github.com/pion/rtp/issues/189

PR #296, opened 2025-02-27
Dont include videoDepacketizer in VP1Packet

jech:
Now that we have separated the Packet/Depacketizer roles for
the AV1 codec, there is no need to include the videoDepacketizer
mixin in AV1Packet.

Related to #291

### DIFF
--- a/codecs/av1_packet.go
+++ b/codecs/av1_packet.go
@@ -136,7 +136,8 @@ type AV1Packet struct {
 	// AV1Frame provides the tools to construct a collection of OBUs from a collection of OBU Elements
 	OBUElements [][]byte
 
-	videoDepacketizer
+	// zeroAllocation prevents populating the OBUElements field
+	zeroAllocation bool
 }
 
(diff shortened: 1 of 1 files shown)


jech:
CC @JoeTurki

JoeTurki (approved):

JoeTurki:
I'll make a proper av1 bitstream and frame reader, and a parser for some OBUs like sq header, but this would fit more in webrtc,

jech:
Let's move h.264 to the same style, so we can remove state from H264Packet and move it into H264Depacketizer.

JoeTurki:
@jech I'm going to start with H264 support is just after AV1, and planning to fix all the reported bugs, and optimize it more, bring it up to the spec more, and will make a H264Depacketizer too  :)

JoeTurki:
@jech Sean added the TWCC to the v4.1.0 milestone I'll help with this after I finish my todo list for this milestone.

PR #298, opened 2025-03-05
Adopt testify/assert for all tests

### LINKED ISSUE #297: Adopt testify for tests
Currently, we don't use any test kit with this package. I think it would be appropriate to use Testify, like we do in other repositories, especially since we're adding more features to this package.
I can do this project.

@Sean-Der what do you think?

JoeTurki:
#### Description

Use testify's assert package instead of the standard library's testing package.

https://github.com/pion/.goassets/pull/222

#### Reference issue
resolve #297

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -19,12 +19,16 @@ linters-settings:
             recommendations:
               - errors
   forbidigo:
+    analyze-types: true
     forbid:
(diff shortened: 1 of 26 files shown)


Sean-Der:
Screenshot 2025-03-04 at 4 56 53 PM

Love that red diff :)

PR #301, opened 2025-03-10
Fix nalu parsing for 3-byte & 4-byte start seqs

rileym-arcules:
#### Description
There is an issue in the current `emitNalu` function where the implementation can skip over 3-byte start sequences. 

The `emitNalu` function was changed in commit bfe92b9 which introduced this regression: https://github.com/pion/rtp/commit/bfe92b95ae6f7195e58eb777dc1d860e511216fb#diff-5eca440e933baa4c5491dcd9ccc7c1f4d807994e6c5b54c304228c34fcc3c963L40 

From analysis of the current `emitNalu` function,
https://github.com/pion/rtp/blob/ee5524bed13b5f257ae7083ba4923001b59dfa59/codecs/h264_packet.go#L45-L67
one can observe that NALUs with 3-byte start sequences may be skipped over if it is the first NALU, or if it is between two NALUs with 4-byte start sequences (in this case: at L52 `end != -1`)

This PR aims to resolve this issue by searching for 3-byte start sequences and accounting for 4-byte sequences afterward.

### DIFF
--- a/codecs/h264_packet_test.go
+++ b/codecs/h264_packet_test.go
@@ -12,6 +12,12 @@ func TestH264Payloader_Payload(t *testing.T) { //nolint:cyclop
 	pck := H264Payloader{}
 	smallpayload := []byte{0x90, 0x90, 0x90}
 	multiplepayload := []byte{0x00, 0x00, 0x01, 0x90, 0x00, 0x00, 0x01, 0x90}
+	mixednalupayload := []byte{
+		0x00, 0x00, 0x01, 0x90,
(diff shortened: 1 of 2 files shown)


codecs/h264_packet_test.go
```diff
@@ -303,28 +315,3 @@ func TestH264Payloader_Payload_SPS_and_PPS_handling(t *testing.T) {
 		t.Fatal("SPS and PPS aren't packed together")
 	}
 }
-
```
rileym-arcules, line 306: Whoops, not intended. Restoring this

PR #305, opened 2025-04-25
Add NewPacketizerWithOptions for configuration

### LINKED ISSUE #111: Allow the ability to define the initial timestamp for the packetizer.
## Summary

Currently the timestamp is defined using a random number

https://github.com/pion/rtp/blob/319e780b743768216fcd1facedd8b26ca208ff01/packetizer.go#L40

This doesn't allow for syncing tracks from multiple sources where we can then swap the tracks from the server without repair.

## Motivation

I have 4 video sources with different quality settings but they cannot be used with ABR since the initial timestamps do not correspond with each other.

JoeTurki:
#### Description
Add NewPacketizerWithOptions to create and configure packetizers, mainly to make it possible to create packetizers with custom timestamp, And for extra options in the future.

Also removed `ssrc` and `PayloadType` from the arguments, we don't use it.

This probably should replace `NewPacketizer` in the new major release.

#### Reference issue
Fixes #111

### DIFF
--- a/packetizer.go
+++ b/packetizer.go
@@ -59,6 +59,51 @@ func NewPacketizer(
 	}
 }
 
+// WithSSRC sets the SSRC for the Packetizer.
+func WithSSRC(ssrc uint32) func(*packetizer) {
+	return func(p *packetizer) {
+		p.SSRC = ssrc
+	}
+}
+
+// WithPayloadType sets the PayloadType for the Packetizer.
+func WithPayloadType(pt uint8) func(*packetizer) {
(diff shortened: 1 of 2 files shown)


JoeTurki:
@Sean-Der this has been in my mind for a while, I'm not sure if this is the right way to do it or wait for the new major release!

Sean-Der:
Go for it! Much nicer not having all the args :)

Sean-Der (approved):

sylwek78:
@roblofthouse

PR #307, opened 2025-05-04
Moved PaddingSize field from Packet to Header

sirzooro:
Parts of pion code passes RTP header and payload via separate parameters instead of using Packet struct to pass them together. As a result value of padding size field may be lost. This change adds PaddingSize field to the Header struct and marks one in Packet as a deprecated.

This is part of fix for https://github.com/pion/webrtc/issues/2403 . Another change is needed for pion/srtp to resolve it fully.


packet.go
```diff
@@ -490,7 +500,11 @@ func (p Packet) Marshal() (buf []byte, err error) {
 
 // MarshalTo serializes the packet and writes to the buffer.
 func (p *Packet) MarshalTo(buf []byte) (n int, err error) {
-	if p.Header.Padding && p.PaddingSize == 0 {
+	if p.PaddingSize > 0 && p.Header.PaddingSize == 0 {
```
JoeTurki, line 503: Why did we remove the `p.Header.Padding` check?

packet.go
```diff
@@ -499,34 +513,47 @@ func (p *Packet) MarshalTo(buf []byte) (n int, err error) {
...
-	return p.Header.MarshalSize() + len(p.Payload) + int(p.PaddingSize)
+	if p.PaddingSize > 0 && p.Header.PaddingSize == 0 {
+		p.Header.PaddingSize = p.PaddingSize
+	}
+
+	return p.Header.MarshalSize() + len(p.Payload) + int(p.Header.PaddingSize)
 }
 
 // Clone returns a deep copy of p.
 func (p Packet) Clone() *Packet {
```
JoeTurki, line 544: Maybe we need to update `Header.Clone()` too?

packet_test.go
```diff
@@ -37,9 +37,9 @@ func TestBasic(t *testing.T) { // nolint:maintidx,cyclop
 			Timestamp:      3653407706,
 			SSRC:           476325762,
 			CSRC:           []uint32{},
+			PaddingSize:    0,
 		},
-		Payload:     rawPkt[20:],
-		PaddingSize: 0,
```
JoeTurki, line 42: Until we remove PaddingSize from the packet, In the next major release, I think we need to keep these tests and make new ones for the better header based logic, Just to insure we don't break something for someone who rely on this logic, what do you think?

JoeTurki (commented):
Thank you for taking care of this project, This something that was on my todo list, I think adding it to the header is the best approach, just few nitpicks

packet.go
sirzooro: This check is in line 507 below. In this and few other functions here I have added code like this one to copy value of PaddingSize field from Packet to Header.

packet.go
sirzooro: `Header.Clone` is fine. These changes here copy value from old to the new field.

JoeTurki: Makes sense!

aler9:
Hello, this change doesn't fix anything inside this specific library (pion/rtp) and makes it more difficult to use since the resulting public interface is less clear and doesn't correspond to a real RTP packet structure anymore, since the padding is moved inside the header while it doesn't belong there (as written inside the code too).

Personally i would have fixed things inside pion/webrtc in order to route `PaddingSize` together with `Header`, creating there (not here) a dedicated structure, like:

```go
type headerPaddingSizePair struct {
    header rtp.Header
    paddingSize int
}
```

instead of making this, that feels like a workaround.

Said that, if there are no alternatives for fixing an important issue, this can be merged since it doesn't cause any conflict with existing code.

JoeTurki:
> instead of making this, that feels like a workaround.

Yes this is a workaround because we can't change `WriteRTP(header *rtp.Header, payload []byte)`, It's true that it's not related to RTP, If we went to do it right, I suggest we fix it in webrtc and srtp and break changes or introduce new APIs, But I think it's safer to do a workaround and wait for pion 5.

We can make a list for the stuff we should fix / break for pion 5, This seems like a good fit!

sirzooro:
> Hello, this change doesn't fix anything inside this specific library (pion/rtp) [...]

Yes. Full fix also requires change to pion/srtp, where padding has to be appended to payload before encryption. pion/srtp has to know padding size to do this. Public API has to be backward compatible, so this looked like a best approach.

> [...] and makes it more difficult to use since the resulting public interface is less clear and doesn't correspond to a real RTP packet structure anymore, since the padding is moved inside the header while it doesn't belong there (as written inside the code too).

Size fields usually are in protocol headers. But RTP was designed to be as compact as possible, so designers did not want to waste extra bytes to send packet size and padding size in RTP header. Using one bit to mark presence of padding and sending padding size in last byte of the packet allowed to reduce packet size, even that this approach is a bit hacky.

> Personally i would have fixed things inside pion/webrtc in order to route `PaddingSize` together with `Header`, creating there (not here) a dedicated structure, like:
> 
> ```go
> type headerPaddingSizePair struct {
>     header rtp.Header
>     paddingSize int
> }
> ```
> 
> instead of making this, that feels like a workaround.
> 
> Said that, if there are no alternatives for fixing an important issue, this can be merged since it doesn't cause any conflict with existing code.

This would break public API, so this is something to consider for pion v5.0.x. I think that it should be changed to send RTP header and raw (marshaled) RTP packet in one struct. SRTP ciphers works with buffers, so having RTP packet as raw bytes would simplify things there. This is especially important for Cryptex (full header encryption, RFC 9335) support. There is already WebRTC extension defined for it, so after some time it will became part of the standard and browsers will start using it: https://w3c.github.io/webrtc-extensions/#rtp-header-extension-encryption

BTW, I am going to implement Cryptex in pion/srtp.

packet.go
```diff
@@ -499,34 +513,47 @@ func (p *Packet) MarshalTo(buf []byte) (n int, err error) {
...
 
-	return n + m + int(p.PaddingSize), nil
+	return offset + m + int(header.PaddingSize), nil
 }
 
 // MarshalSize returns the size of the packet once marshaled.
 func (p Packet) MarshalSize() int {
-	return p.Header.MarshalSize() + len(p.Payload) + int(p.PaddingSize)
+	if p.PaddingSize > 0 && p.Header.PaddingSize == 0 {
+		p.Header.PaddingSize = p.PaddingSize
```
at-wat, line 537: The function name looks like an immutable method but have a side effect. (`Clone()` as well)
I think it's reasonable for now but maybe better to note this as a godoc comment.

at-wat (approved):
LGTM!
Let's have a full fix in the next major update

packet.go
sirzooro: Thanks. I have updated code to remove this side effect.

sirzooro:
Thanks! To fully fix this, I need to create two more PRs, for srtp and webtc.

PR #314, opened 2025-06-30
Return packaging bytes for h265 unmarshal

cnderrauber:
Return packaging bytes for h265 unmarshal, assemble the fragmented packets.

### DIFF
--- a/codecs/h265_packet.go
+++ b/codecs/h265_packet.go
@@ -15,8 +15,9 @@ import (
 //
 
 var (
-	errH265CorruptedPacket   = errors.New("corrupted h265 packet")
-	errInvalidH265PacketType = errors.New("invalid h265 packet type")
+	errH265CorruptedPacket          = errors.New("corrupted h265 packet")
+	errInvalidH265PacketType        = errors.New("invalid h265 packet type")
+	errExpectFragmentationStartUnit = errors.New("expecting a fragmentation start unit")
 )
 
 //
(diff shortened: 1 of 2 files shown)


codecs/h265_packet.go
```diff
@@ -536,6 +559,63 @@ func (p *H265FragmentationUnitPacket) Payload() []byte {
 
 func (p *H265FragmentationUnitPacket) isH265Packet() {}
 
+// H265FragmentationPacket represents a Fragmentation packet, which contains one or more Fragmentation Units.
+type H265FragmentationPacket struct {
```
JoeTurki, line 565: Is there a reason to make `H265FragmentationPacket` and its members public?

JoeTurki (commented):
I was working on fixing HEVC support, But i found the parser and packetizer to be wrong and inefficient and have many mistakes when it comes to the spec, So I'm opting to rewrite them soon. https://github.com/pion/webrtc/issues/3137

I think this can be shipped as a temporary fix.

codecs/h265_packet.go
cnderrauber: It is not required to be public in my case, keep it consistent with other Packet(s).

codecs/h265_packet.go
```diff
@@ -691,6 +774,21 @@ func (p *H265PACIPacket) Unmarshal(payload []byte) ([]byte, error) {
 
 func (p *H265PACIPacket) isH265Packet() {}
 
+func (p *H265PACIPacket) doPackaging(buf []byte) []byte {
+	buf = append(buf, annexbNALUStartCode...)
+	buf = append(buf, byte(p.payloadHeader>>8), byte(p.payloadHeader&0xFF))
+
+	binary.BigEndian.AppendUint16(buf, p.paciHeaderFields)
```
JoeTurki, line 781: Did you forgot re-assigning?

boks1971 (approved):
Don't know the details of the bitstream format, but code lgtm!

If @JoeTurki is good with this as interim fix, sgtm!

codecs/h265_packet.go
JoeTurki: Makes sense, I think the original HEVC implementation is very noisy, we'll have to move and hide most of this abstraction in v2 anyway.

PR #343, opened 2025-12-03
use atomic instead of lock in sequencer

AnshulMalik:


Sean-Der:
Nice optimization @AnshulMalik!

Are you profiling the entire code/lock contention? What else have you found?

Sean-Der (approved):

AnshulMalik:
I am trying it in my free time. Most of the code is already pretty efficient, just finding small things.

PR #360, opened 2026-05-01
Add method to clear header extensions

boks1971:
Adding a method to clear RTP header extensions.
Also, modifying `DelExtension` to clear the extension flag if DelExtension will delete the last extension.

### DIFF
--- a/packet.go
+++ b/packet.go
@@ -503,12 +503,23 @@ func (h *Header) DelExtension(id uint8) error {
 	for i, extension := range h.Extensions {
 		if extension.id == id {
 			h.Extensions = append(h.Extensions[:i], h.Extensions[i+1:]...)
+			if len(h.Extensions) == 0 {
+				h.Extension = false
+				h.ExtensionProfile = 0
+			}
 
 			return nil
 		}
 	}
 
 	return errHeaderExtensionNotFound
+}
+
+// ClearExtensions Removes all RTP Header extension.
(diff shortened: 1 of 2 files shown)


JoTurk (approved):
Thank you, I wonder what is your use case?

packet.go
```diff
@@ -503,6 +503,10 @@ func (h *Header) DelExtension(id uint8) error {
 	for i, extension := range h.Extensions {
 		if extension.id == id {
 			h.Extensions = append(h.Extensions[:i], h.Extensions[i+1:]...)
+			if len(h.Extensions) == 0 {
+				h.Extension = false
+				h.ExtensionProfile = 0
+			}
```
JoTurk, line 509: good catch, would be cool to include a unit test for Marshal to make sure we don't set the extension bit.

boks1971:
> Thank you, I wonder what is your use case?

Exporting to some thing like gstreamer pipeline where the extensions are not useful. Just need the payload.

packet.go
boks1971: Will have a look.

boks1971: Added [here](https://github.com/pion/rtp/pull/360/changes/fdcdb80949141cd3f234ca339cf39bdcbea08396) @JoTurk

JoTurk: Thank you.

Issue #361, opened 2026-05-23
MIDI RTP (RFC 6295)


jwetzell:
There is a MIDI payload type for RTP ([RFC 6295](https://datatracker.ietf.org/doc/html/rfc6295)). I'm still making my way through the layout of this library and might be able to give an implementation of this a go.

JoTurk:
Hi @jwetzell 

This library is intended for webrtc we usually support payload formats and extensions that are supported by usual browser webrtc implementations. But we plan to change this by the next major release, since RTCRTPTransport is likely to become a thing in browsers, and we can use this library in experiments as RTP over Quic or over SCTP.

But we plan to refactor this library first and standardize interfaces more, and a map for payloaders and depacketizers.

jwetzell:
I'm seeing that trend amongst many of Pion's libraries. Hopefully it can become general RTP library that isn't coupled to a specific use-case.

Sean-Der:
Thanks for the feedback @jwetzell, what’s the trend?

* Staying WebRTC specific?
* being in ‘refactor mode’?

Would love to make Pion as easy/exciting as possible to contribute too :)

JoTurk:
Also we can't support this rfc before https://github.com/pion/rtp/issues/69 is resolved by changing the payloaders interface so they have direct access to the rtp header, because this
Payload format has use for the m bit, it's a good rfc to base the new design agains if we want to make this library not webrtc specific.
Currently the vvc and hevc payloaders are missing some optional features because of this too.

JoTurk:
@Sean-Der I think @jwetzell is talking about the trend of making pion libraries general use, they joined to help with stuff with mdns that aren't related to webrtc, context https://github.com/pion/mdns/issues/253 we currently don't have any libraries on refactor mode other than dtls (for 1.3, and currently we have a release branch and a main branch), there are planned refactors for ice and rtp, but the work didn't start yet and we still merge PRs to both of them.

I want your thought on making Pion libraries not webrtc-centeric there is some disagreements example https://github.com/pion/sdp/pull/217#issuecomment-3315570837 I'm 50/50 on this and  I would love your take to use it as a reference and for future decisions, thank you.

jwetzell:
Yes @Sean-Der, @JoTurk is correct. I completely understand that the role of a lot of these ancillary libraries is/was to support the main WebRTC "goal", but people aren't always exposed to them through that context (`pion/mdns` is a top result when searching "golang mDNS").

[...]

PR #362, opened 2026-05-25
Add RTP/v3 interface design

### LINKED ISSUE #69: Always set Marker bit to false?
Packetize always sets the Marker bit for the last payload in a sequence. I don't believe that that is semantically appropriate. The marker bit is defined to be application- (codec-)specific.

For example, for voice codecs, [RFC 3551 section 4.1 says](https://tools.ietf.org/html/rfc3551#section-4.1):

> For applications which send either no packets or occasional comfort-noise packets during silence, the first packet of a talkspurt, that is, the first packet after a silence period during which packets have not been transmitted contiguously, SHOULD be distinguished by setting the marker bit in the RTP data header to one.  The marker bit in all other packets is zero.  The beginning of a talkspurt MAY be used to adjust the playout delay to reflect changing network delays. Applications without silence suppression MUST set the marker bit to zero.

If you feed a sequence of pre-constructed media samples in, as in https://github.com/Sean-Der/audio-spew/blob/687cf84003383d5a9ced1384264d9b99f669cfe5/main.go#L88-L100, every single packet gets the Marker bit set, which is definitely not right.

I suspect that the correct thing for this package to do is to never set the Marker bit, leaving it to the parent application to set it as needed.

cc @Sean-Der

### LINKED ISSUE #70: Add a way to set codec specific payload descriptor flags
For example, 
https://github.com/pion/rtp/blob/3ea35019945910320f917f2aef8a9fbc4ace4d83/codecs/vp8_packet.go#L67-L72
N flag can be set to indicate the frame can be discarded on packet loss. (The value for each frame is given by the encoder.)

JoTurk:
#### Description
Note: that this is V3 because rtp/v2 was released by mistake a few years ago, This moves the library more towards a generic RTP library, this is backward compatible with v1 for users with custom packetizers/depacketizers, but we should drop this in v4.

Today the payloader interface is:
```go
type Payloader interface {
    Payload(mtu uint16, payload []byte) [][]byte
}
```

and the generic webrtc-centeric packetizer builds RTP packets afterward. In the current interface, the market bit is hardcoded as "true on the last emitted payload fragment" which was true for many webrtc rtp payloads.
 
However, RTP defines the market bit as profile-specific not last fragment, RTP-MIDI is a clean example of the limitation of this interface #361 This RFC says the marker bit depends on the media type, for mpeg-genric for example it's always one, and this can't be expressed by "last payload wins".

When it comes to the depacketizer, the API has related problems:

```go
type Depacketizer interface {
    Unmarshal(packet []byte) ([]byte, error)
    IsPartitionHead(payload []byte) bool
    IsPartitionTail(marker bool, payload []byte) bool
}
```
Unmarshal only receives the RTP payload, IsPartitionHead receives only payload bytes, and only IsPartitionTail gets the marker bit. The full RTP header, timestamp, sequence number, extensions, padding state, SSRC, and payload type are not available to the depacketizer methods.

Then, The samplebuilder then calls the depacketizer in a way that reinforces this limitation: it checks the tail with Marker and Payload, checks the head with only Payload, and appends the result of Unmarshal(payload) to the sample data.

This broke down for me when I was finishing av1 support because RTP can omit the length field, and had to make the interface stateful to deal with stuff like W !=0, and it wasn't not possible to honor the push / append model.

Many depacketizers, carry mutable depacketization state such as fuaBuffer in H264Packet, so a packet type is also acting as a stateful depacketizer.

this refactors the interface into

```
codec bitstream reader/writer
        ↓
RTP payload-format packetizer/depacketizer
        ↓
generic RTP packet sequencer/parser
        ↓
sample builder / jitter buffer / transport integration
```
which let's us define interfaces for codecs that have different input payloads e.g accepting avc and annex-b h264 without having hacks or random options, Also being able to write avc or annex-b as output without multiple transformation.

#### Reference issue
Fixes #69 and fixes #70

### DIFF
--- a/payload_format.go
+++ b/payload_format.go
@@ -1,0 +1,159 @@
+// SPDX-FileCopyrightText: 2026 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+package rtp
+
+// MediaFormat describes the bitstream or sample representation passed to a payload format.
+type MediaFormat string
+
+// PacketizeContext contains the RTP state and negotiated parameters available
+// to a payload-format packetizer.
+type PacketizeContext struct {
+	MTU         int
+	PayloadType uint8
+	SSRC        uint32
+	Timestamp   uint32
+	Sequencer   Sequencer
+	Params      any
+}
+
+// MediaSample is the media input passed to a payload-format packetizer.
+type MediaSample struct {
+	Payload  []byte
+	Duration uint32
+	Format   MediaFormat
+	Metadata any
+}
+
+// ExtensionWrite applies a payload-format-owned RTP header extension.
+type ExtensionWrite interface {
+	Apply(*Header) error
+}
+
+// PayloadFragment is an RTP payload fragment plus the RTP header semantics
+// owned by the payload format.
+type PayloadFragment struct {
+	Payload         []byte
+	Marker          bool
+	TimestampOffset uint32
+	Extensions      []ExtensionWrite
+	MutateHeader    func(*Header) error
+}
+
+// PayloadFormatPacketizer packetizes media samples into RTP payload fragments.
+type PayloadFormatPacketizer interface {
+	Packetize(ctx PacketizeContext, sample MediaSample, emit func(PayloadFragment) error) error
+	Flush(ctx PacketizeContext, emit func(PayloadFragment) error) error
+	Reset()
+}
+
+// PacketView exposes the RTP header and payload to payload-format depacketizers.
+type PacketView struct {
+	Header  *Header
+	Payload []byte
+}
+
+// PacketInfo describes payload-format sample boundaries and parsed packet metadata.
+type PacketInfo struct {
+	StartsSample bool
+	EndsSample   bool
+	KeyFrame     bool
+	Metadata     any
+}
+
+// PayloadFormatDepacketizer inspects RTP packets and appends their media bytes
+// to a sample being assembled.
+type PayloadFormatDepacketizer interface {
+	Inspect(packet PacketView) (PacketInfo, error)
+	AppendToSample(dst []byte, packet PacketView) ([]byte, error)
+	Reset()
+}
+
+// LegacyPayloaderAdapter adapts a legacy Payloader to PayloadFormatPacketizer.
+type LegacyPayloaderAdapter struct {
+	Payloader Payloader
+}
+
+// Packetize adapts the legacy Payloader Payload method to PayloadFormatPacketizer.
+func (a LegacyPayloaderAdapter) Packetize(
+	ctx PacketizeContext,
+	sample MediaSample,
+	emit func(PayloadFragment) error,
+) error {
+	payloads := a.Payloader.Payload(legacyPayloaderMTU(ctx.MTU), sample.Payload)
+	for i, payload := range payloads {
+		if err := emit(PayloadFragment{
+			Payload: payload,
+			Marker:  i == len(payloads)-1,
+		}); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Flush adapts legacy payloaders, which do not buffer pending payload fragments.
+func (a LegacyPayloaderAdapter) Flush(_ PacketizeContext, _ func(PayloadFragment) error) error {
+	return nil
+}
+
+// Reset adapts legacy payloaders, which do not expose reset behavior.
+func (a LegacyPayloaderAdapter) Reset() {}
+
+// LegacyDepacketizerAdapter adapts a legacy Depacketizer to PayloadFormatDepacketizer.
+type LegacyDepacketizerAdapter struct {
+	Depacketizer Depacketizer
+}
+
+// Inspect adapts the legacy partition boundary methods to PayloadFormatDepacketizer.
+func (a LegacyDepacketizerAdapter) Inspect(packet PacketView) (PacketInfo, error) {
+	marker := false
+	if packet.Header != nil {
+		marker = packet.Header.Marker
+	}
+
+	return PacketInfo{
+		StartsSample: a.Depacketizer.IsPartitionHead(packet.Payload),
+		EndsSample:   a.Depacketizer.IsPartitionTail(marker, packet.Payload),
+	}, nil
+}
+
+// AppendToSample adapts the legacy Unmarshal method to PayloadFormatDepacketizer.
+func (a LegacyDepacketizerAdapter) AppendToSample(dst []byte, packet PacketView) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


JoTurk:
(I submitted without writing the description, writing it with edit ig.)

JoTurk:
@jech you didn't like the current RTP interface, what do you think about this one?

sirzooro:
There are many things marked as deprecated in pion/rtp, it would be good to remove them during migration to v3.

JoTurk:
@sirzooro thank you for the reminder, and yes this won't be the last PR, we should also move some stuff around.

payload_format.go
```diff
@@ -0,0 +1,159 @@
+// SPDX-FileCopyrightText: 2026 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+package rtp
+
+// MediaFormat describes the bitstream or sample representation passed to a payload format.
```
jwetzell, line 6: Not sure how much is applicable here, but I think whenever possible the RFC should be referenced for naming and rational behind particular concepts. Even if its just a comment with the RFC # + section.

payload_format.go
```diff
@@ -0,0 +1,159 @@
...
+	Header  *Header
+	Payload []byte
+}
+
+// PacketInfo describes payload-format sample boundaries and parsed packet metadata.
+type PacketInfo struct {
+	StartsSample bool
+	EndsSample   bool
+	KeyFrame     bool
+	Metadata     any
```
jwetzell, line 61: Where are the fields here coming from? Are these something all RTP packets share?

JoTurk: This essintally moving pion's webrtc sample writer boundaries to rtp, so rtp can control what a video or audio sample is, e.g in av1 this will be a full av1 frame or video sequence, this is really important for some payloads like implementing correct DONL for hevc, currently the sample bounder is weak.

payload_format.go
JoTurk: This is mostly application logic but I'll try to reference all my reasons, and specs behind every interface.

payload_format.go
jwetzell: So every `depacketizer` would have to deal with the concepts of samples and keyframes even if they don't make sense for that particular payload format (probably just by ignoring the fields)?

jech:
I need to think it over, but at first sight this looks good.  Two questions.

Why do we need the PacketizerContext as an explicit parameter?  The Packetizer is a stateful object, right, so why not make the context part of the Packetizer?

Why do we need the Flush method?

payload_format.go
JoTurk: Yes but we can have something like a base - depackitzer / soft inheritance for payloads that don't need explicit access, most do btw, we just weren't so explicit with our implementions!

JoTurk:
@jech yeah I agree it was the first iteration and I just added too much, going to remove context from the public input, the user only needs to configure de/packetizers once (e.g avc vs annexb for h264).
Also agree about flush it's not needed since the new interface owns all the logic.
Going to remove them both.

jech:
>
> ***@***.**** commented on this pull request.
>
>
>
> So every *depacketizer* would have to deal with the concepts of samples 
> and keyframes even if they don't make sense for that particular payload 
> format (probably just by ignoring the fields)?

As I understand it, a sample corresponds to a sequence of rtp packets 
that have the same timestamp. This corresponds to a frame in Opus and 
VP8, and a Picture in VP9. I forget the AV1 term.

A keyframe is just a sample at which you may start decoding a stream 
without any knowledge beyond what is in the SDP.  The concept maps to 
every current codec (or else how would you join a conference in the 
middle?), although not always in an obvious manger (think of the 
parameter OBUs in AVC, for example).

Joe, shall we rename sample to frame? Too bad for VP9, but I think they 
got the terminology wrong.

fippo:
A keyframe is a frame with no dependencies to prior frames. https://medooze.medium.com/mastering-the-av1-svc-chains-a4b2a6a23925 is still a very good reference. RTP shouldn't know about this ideally.

JoTurk:
@fippo In av1, RTP needs to know if it's a keyframe to set the N bit correctly, this is a limitation in Pion/RTP AV1 impl is that we set the N bit for all seq-headers and not only for key frames, because I didn't want to parse again in the packetizer https://github.com/pion/rtp/blob/1edc72d15cbeb6ffe0fe40f863797273c19083f2/codecs/av1_packet_test.go#L1447-L1490 Libwebrtc sets it correctly to only keyframes https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/rtp_rtcp/source/rtp_packetizer_av1.cc;l=377?q=rtp_packetizer_av1.cc&ss=chromium%2Fchromium%2Fsrc 


While it seems like setting the N flag for all seq headers doesn't break any impls, e.g chrome doesn't even construct full / real frames, because decoders don't require them.  https://chromium.googlesource.com/external/webrtc/+/HEAD/modules/rtp_rtcp/source/video_rtp_depacketizer_av1.cc#387

But this change will make us correct Pion AV1 and other codecs impls that needs to be aware about "key frame".


A similar limitation is that we don't set the P flag at all in our hevc impl, and we should set it at the end of video seq, This redesign should make all of this possible without parsing again inside RTP (and pion/webrtc and in the user application sometimes e.g with the current design the user has to do avc to annex-b before passing it to pion!).

jech:
> A keyframe is a frame with no dependencies to prior frames. https://medooze.medium.com/mastering-the-av1-svc-chains-a4b2a6a23925 is still a very good reference. RTP shouldn't know about this ideally.

In addition to what Joe said, applications need to be able to identify a keyframe.  For example, when a client joins a conference, it must not try to decode the stream until it sees a keyframe.

jwetzell:
@JoTurk I think I'm piecing this together so the `PacketInfo` here would kind of in the realm of [this](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/rtp_rtcp/source/rtp_packet_to_send.h) in the chromium example?

payload_format.go
```diff
@@ -0,0 +1,159 @@
...
+	MTU         int
+	PayloadType uint8
+	SSRC        uint32
+	Timestamp   uint32
+	Sequencer   Sequencer
+	Params      any
+}
+
+// MediaSample is the media input passed to a payload-format packetizer.
+type MediaSample struct {
```
sirzooro, line 21: There is no timestamp field here. It would would be needed to implement extensions like ones from RFC 6051, or deal with gaps in media stream.

sirzooro:
It would be good to also return error from various `OneByteHeaderExtension`, `TwoByteHeaderExtension` and `RawExtension` methods. Now they have no way to signal problem to the caller.

PR #363, opened 2026-06-21
Check data length it RTP ExtHdr handlers

sirzooro:
Added data length checks to OneByteHeaderExtension, TwoByteHeaderExtension and RawExtension to prevent panics. They are probably not used by pion itself (did not check all packages here), so fix is just in case someone else still uses them.

### DIFF
--- a/header_extension.go
+++ b/header_extension.go
@@ -10,6 +10,18 @@ import (
 	"io"
 )
 
+/*
+  RFC 3550, RTP Header Extension:
+
+    0                   1                   2                   3
+    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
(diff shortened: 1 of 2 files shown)


sirzooro:
@JoTurk do we still need `OneByteHeaderExtension`, `TwoByteHeaderExtension` and `RawExtension`? I wonder if we could mark them as deprecated and remove in next major version.

header_extension.go
```diff
@@ -70,6 +86,9 @@ func (e *OneByteHeaderExtension) Set(id uint8, buf []byte) error {
 
 // GetIDs returns the available IDs.
 func (e *OneByteHeaderExtension) GetIDs() []uint8 {
+	if len(e.payload) < 4 {
+		return nil
+	}
```
JoTurk, line 91: It's a shame that we can't return an error.

JoTurk (approved):
I see many issues with these extensions but they are all pre existing sorry for taking time with the review.
I think we shouldn't remove them either since we're making this library a genetic rtp library

header_extension.go
sirzooro: Yeah. We could change this in RTP/v3. I will add comment there.

PR #364, opened 2026-07-02
Skip allocations when reusing VP9 packets

kcaffrey:
If a caller reuses `codecs.VP9Packet` (after resetting the state appropriately), then reuse already allocated slices if they are of sufficient length instead of unconditionally allocating. Add tests to verify that reuse is safe and is zero allocations.

### DIFF
--- a/codecs/vp9_packet.go
+++ b/codecs/vp9_packet.go
@@ -452,8 +452,8 @@ func (p *VP9Packet) parseSSData(packet []byte, pos int) (int, error) { // nolint
 	p.NG = 0
 
 	if p.Y {
-		p.Width = make([]uint16, NS)
-		p.Height = make([]uint16, NS)
+		p.Width = resizeUint16Slice(p.Width, int(NS))
+		p.Height = resizeUint16Slice(p.Height, int(NS))
 		for i := 0; i < int(NS); i++ {
 			if len(packet) <= (pos + 3) {
 				return pos, errShortPacket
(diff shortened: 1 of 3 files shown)


codecs/vp9_packet.go
```diff
@@ -451,9 +451,17 @@ func (p *VP9Packet) parseSSData(packet []byte, pos int) (int, error) { // nolint
...
+		if cap(p.Width) < int(NS) {
+			p.Width = make([]uint16, NS)
+		} else {
+			p.Width = p.Width[:NS]
+		}
+		if cap(p.Height) < int(NS) {
+			p.Height = make([]uint16, NS)
+		} else {
+			p.Height = p.Height[:NS]
+		}
```
JoTurk: Nit: instead of adding a nolint i suggest you refactor this into a small helper resizeUint16Slice or something.

codecs/vp9_packet.go
```diff
@@ -482,10 +490,15 @@ func (p *VP9Packet) parseSSData(packet []byte, pos int) (int, error) { // nolint
 
 		p.PGTID = append(p.PGTID, packet[pos]>>5)
 		p.PGU = append(p.PGU, packet[pos]&0x10 != 0)
-		R := (packet[pos] >> 2) & 0x3
+		R := (packet[pos] >> 2) & 0x3 // nolint: varnamelen
```
JoTurk: Can you rename this to reference or something instead of adding a nolint.

JoTurk (commented):
Change looks good, thank you.
just two nits.

codecs/vp9_packet.go
kcaffrey: Good call, done

codecs/vp9_packet.go
kcaffrey: Done

JoTurk (approved):
thank you.

kcaffrey:
Thanks for the review @JoTurk. Any chance you can create a new release with this? I saw the workflow for releasing but I don't have access to run it.

JoTurk:
@kcaffrey you do have access. Just tag v1.10.3 off main and push it and it will create the release :)

PR #366, opened 2026-07-18
Support multiple sps/pps/vps in h265 AU

JoTurk:
#### Description
As discussed in discord, the SkipAggregation behavior isn't correct for H265 because streams can have multiple sps/pps/vps with different ids. while I wasn't able to find any non-svc real media files with this it's correct in the spec and chrome can play even tho chrome doesn't support SHVC http://download.tsi.telecom-paristech.fr/gpac/MPEG/LHEVCFF/Conformance/shvc_hev1_single_track.mp4 it's able to play this file after the fix (tested with hardware decode NVENC).
The problem is that the order in this file is:
```
#id=0  VPS
#id=0  SPS  
#id=0  PPS
#id=1  SPS  
#id=1  PPS
#id=0  IDR 
#id=1  IDR
```
chrome picks layer #0 and because we cache the last SPS/PPS the file doesn't play before this fix. I'm not sure if we should merge this as this sounds like a niche with more complexity trade off but it might become more common with svc h265 or other uses and it can save someone a few hours of debug in the future.
This is only a draft because i didn't add tests yet

### DIFF
--- a/codecs/h265_packet.go
+++ b/codecs/h265_packet.go
@@ -1726,19 +1726,78 @@ func (p *H265Packet) Packet() isH265Packet {
 	return p.packet
 }
 
+type h265ParamSetEntry struct {
+	typ    uint8
+	id     uint32
+	packet h265SingleNALUnitPacket
+}
+
+// h265ParamSetCache holds the latest VPS/SPS/PPS for each type and parameter-set ID,
+// ordered by the latest occurrence of each parameter set in the bitstream.
+type h265ParamSetCache struct {
+	entries []h265ParamSetEntry
+}
+
 // H265Payloader payloads H265 packets.
 type H265Payloader struct {
 	// Deprecated: Has no effect.
 	AddDONL bool
 	// SkipAggregation disables H265 aggregation packets.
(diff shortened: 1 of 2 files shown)


Ibn-Butt-Toota (approved):
Maybe we can create a section in the readme to document spec-breaking implementations or conflicts like this so it can be easier for observers to find? Or maybe if it's feasible we could add an experimental flag...

JoTurk:
@Ibn-Butt-Toota thank you for the review, this is the spec correct way to cache sps/pps/vps the problem is that it's not clear if multiple sps/pps/vps are used outside of shvc and if it's common :)

Ibn-Butt-Toota:
Sorry I misread the main comment. Thank you for clarifying. The PR still looks good (I know it's pending tests still).

Sean-Der (approved):
oh hell yea, thank you for fixing this @JoTurk !

JoTurk:
@Sean-Der thank you for the review <3 turned out there were many other edge cases with multiple sps/pps/vps idk why h265 is this complex, hopefully I handled all of them :)

